PR #49, opened 2015-08-09
Don't prepend relative URLs with InputFolder in Less files

Rohansi:
When using [these files](https://gist.github.com/Rohansi/102e509337ac73da70b9), the output CSS is prepended with InputFolder but it shouldn't be.

Reproduced in 0.7.0-beta, tested my changes and the URL stops being modified and imports still work.

### DIFF
--- a/Wyam.Modules.Less/Less.cs
+++ b/Wyam.Modules.Less/Less.cs
@@ -19,7 +19,6 @@ namespace Wyam.Modules.Less
         public IEnumerable<IDocument> Execute(IReadOnlyList<IDocument> inputs, IExecutionContext context)
         {
             DotlessConfiguration config = DotlessConfiguration.GetDefault();
-            config.RootPath = context.InputFolder;
             config.Logger = typeof (LessLogger);
 
             EngineFactory engineFactory = new EngineFactory(config);


daveaglick:
Great catch - and especially thanks for the repro. I think the DotLess `RootPath` is also used for finding included Less files, in which case removing it would break modular Less setups (like Bootstrap). There seems to be an assumption that Less content and other resources are all in the same folder structure (as opposed to the input/output convention Wyam uses). 

I'll have to take a closer look to see if I can figure out how to find included Less files in the input folder while still outputting correct relative paths in the outputs. Hopefully there's an interface or something to supply custom behavior. If not, a global search and replace after processing will probably work. In any case, I'll figure it out.

Thanks again!

Rohansi:
According to the documentation it isn't used as the path for imported less files.
- [dotless.Compiler RootPath option](https://github.com/dotless/dotless/blob/master/src/dotless.Compiler/Program.cs#L269)
- [dotless RootPath option XML doc](https://github.com/dotless/dotless/blob/6936c3bd5fa40357b2dc667c31dc848df3609d8f/src/dotless.Core/configuration/DotlessConfiguration.cs#L124-L128)
- [lessc RootPath option](http://lesscss.org/usage/#command-line-usage-rootpath)
- [dotless Importer.cs, the only real usage of RootPath](https://github.com/dotless/dotless/blob/6936c3bd5fa40357b2dc667c31dc848df3609d8f/src/dotless.Core/Importers/Importer.cs#L332-L347)

dotless [uses CurrentDirectory](https://github.com/dotless/dotless/blob/6936c3bd5fa40357b2dc667c31dc848df3609d8f/src/dotless.Core/Importers/Importer.cs#L247-L300) as the base path for importing other less files, so everything should still work.

I just tested this change on your website and everything is still working. In fact, it's a bit better than the live site because it's affected by the same issue! It's trying to load some glyphicons font from file:///.

daveaglick:
Awesome - thanks for digging into it! I'm merging the PR now.

PR #52, opened 2015-08-14
ImageProcessor module

dodyg:
Implement most of ImageProcessor image library functionalities (Alpha, Brightness, Constrain, Contrast, Filter, Format, Hue, Quality, Resize, Saturation, Tint, Vignette)


daveaglick:
This is truly incredible - thanks for all your hard work on it. Also worth noting that the process exposed a bunch of other shortcomings and ideas, so the act of putting this module together improved the project even more than just the code would suggest.

Unfortunately, I've got a lot of "day job" work today, so I probably won't get a chance to merge it in until tomorrow. When I do, I'll publish a new release - lots of great stuff in this one. Thanks again!

daveaglick:
Thanks again! I've merged this PR. I'm getting some test failures (probably due to moving PathHelper to Wyam.Common) that I need to address. After that, I'll be publishing the new version with image processing support.

dodyg:
Great. There are image optimization functionalities that can be incorporated later. I am rebuilding one of website using wyamio. That should be ready this week.

daveaglick:
Excellent - let me know when the site is done and I'll add it to a new "Community" section on the website that's going to include sites built with Wyam, blog posts, contributors, etc.

Also, unless you have any major objections, I'm going to rename this module to `Image` and the library to `Wyam.Modules.Images`. I think that fits better with the notion that modules should be named based on what they do and not how they do it (for example, `Less` instead of `DotLess`). I also think it makes the fluent interface even friendlier by being able to type `Image().Resize(...).Brighten(...)`, etc.

dodyg:
Good idea. Please do the rename.

PR #75, opened 2015-09-30
Download module 

dodyg:
This module enables download of resources using GET request. It has options to pass request headers as well.


daveaglick:
Woot!

PR #80, opened 2015-10-20
Added Feature to AutoLink Module to only Match howl Words

LokiMidgard:
I had problems with the AutoLink Module, it created links in the middle of a word.

So I added WithMatchOnlyHowlWord() Method and checked before adding the new link if the next char and the previous char is a letter or digit. If not the link will be added, else the node is not regarded as Root.


daveaglick:
Thanks for this! Curious, why "howl word" instead of "whole word"?

LokiMidgard:
Most likely because my English isn't the best.

I fixed it.

daveaglick:
:thumbsup: No problem - I was worried there was some special meaning for the difference that I wasn't aware of. Thanks again, this was a great catch.

PR #81, opened 2015-10-20
Created Module HtmlEscape

LokiMidgard:
I'm using this Module to escape umlauts.
It supports a Blacklist mode and a Whitelist mode. (specifying what you want to escape or what you don't want to escape)


daveaglick:
Another excellent contribution - thanks a lot!

PR #83, opened 2015-10-23
Added Overload for Execute that is only performed once per Pipeline

LokiMidgard:
I needed to execute some code only once in the pipeline.

So I thought overloading Execute was the appropriate way.


daveaglick:
Great idea, thanks!

Issue #72, opened 2015-09-23
Implement themes


daveaglick:
The concept of themes go hand-in-hand with recipes (#1) since a given theme is probably only going to work with a given recipe (since it'll rely on the specific metadata, document layout, etc. in that recipe). That said, the main recipes (like blog, docs, etc.) should each have a set of themes.

A theme should contain one or more files that act as though they are in the `InputFolder` provided a file with the same name isn't _actually_ in the `InputFolder`. This is similar to the [Hugo theme system](https://gohugo.io/themes/customizing/). One idea on how to make this work is to introduce an extra `ThemeFolder`. Whenever files need to be found (Razor layouts, Less includes, etc.), the `InputFolder` will be searched first and then the `ThemeFolder` will be searched. This might require an extra file system abstraction.

This will impact modules because those that read from the file system like `ReadFiles` should probably also use the new abstraction and treat files in the `ThemeFolder` as if they were in the `InputFolder` too. That way, plain resources like CSS or JavaScript will be handled in the normal workflow and can be overridden by placing overrides in the `InputFolder`.

Best idea at the moment for packaging themes is to use content-only NuGet packages (just like we do for dumping resources directly in the `InputFolder`). Not sure what the future holds since content-only is apparently going away (?). These theme packages will somehow have to be specially delineated so that content is copied to the `ThemeFolder` instead of the `InputFolder` like it normally would be. This approach also provides theme authors a little flexibility since they wouldn't have to distribute as a NuGet package. The theme could just be zipped up, distributed, and then unzipped to the `ThemeFolder`.

There should also be a way to specify a theme on the command line using something like a `--theme` flag. This would presumably refer to a NuGet package or maybe a zip file.
- Perhaps first check the working path combined with the given theme name/path (could also provide a full path in the flag), adding a `.zip` extension if one isn't already there. If a file is found, and there isn't already a `ThemeFolder` we'll unzip it to the `ThemeFolder`.
- Then, if no file is found, check for a NuGet package with the given name and if found, extract all contained content to the `ThemeFolder`, but again only if the `ThemeFolder` isn't already there.
- Also provide an `--updateTheme` flag that deletes and recreates the `ThemeFolder` from the specified source (so `--theme` must be provided if `--updateTheme` is used)

daveaglick:
[...]

PR #84, opened 2015-10-23
Git-Module Added

LokiMidgard:
Created two Modules that work with git.
- The first Module search for every file in the pipeline the contributors and add it to the Metadata.
- The second Module creates a page for every contributor. Adding information what commits every user made.

Unfortunatly I'm not sure how to write the tests for the git functionality.

### DIFF
--- a/Wyam.Modules.Git/Autor.cs
+++ b/Wyam.Modules.Git/Autor.cs
@@ -1,0 +1,19 @@
+﻿using LibGit2Sharp;
+using System;
+
+namespace Wyam.Modules.Git
+{
+    public class Autor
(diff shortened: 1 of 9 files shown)


daveaglick:
Another excellent set of modules - keep them coming! Two small comments:
- I've merged, but I'm going to rename `Autor` to `Author` :)
- A note to self: when I get to themes in #72, I'll be implementing an entire IO abstraction layer to replace the one in `System.IO`. When that happens, the use of the LessStupidPath library will need to be replaced with the new abstraction (I'll take care of it - just wanted to make a note).

PR #86, opened 2015-10-24
Created Module to define Metadata on Folder level

LokiMidgard:
This Module allows to specify Metadata on folder level.

Adding a file to a folder with the name "local.metadata" copy's the metadata of this document to any other document in the same folder.

Adding a file to a folder with the name "inherit.metadata" copy's the Metadata of this document to any other document in the same folder and subfolder.

I wasn't sure in which project it could fit so I created a new one with Metadata as name.


daveaglick:
Thanks for merging the upstream changes and updating the PR so quickly :).

I can see where you're going with this, and think it's a cool idea.I just have a few comments:
- Can the module be named DirectoryMeta instead of DirectoryMetadata? That would match the existing Meta module name and help avoid confusion with the `IMetadata` interface (which is why I named the other module Meta).
- Can we also rename `WithMetadata()` to `WithMetadataFile()` and remove `WithinheritMetadata()` entirely? The `WithMetadataFile()` could have an optional parameter `inherited` with a default (of `false`?) and we could then specify as many file names as we wanted. In fact, maybe `WithOverride()` could also be removed and made an optional per-file parameter for override could be added. The signatures would then look like `WithMetadataFile(string metadataFileName, bool inherited = false, bool override = false)` and `WithMetadataFile(DocumentConfig metadataFileName, bool inherited = false, bool override = false)`.
- We'll have to be careful to document that this module assumes the `IDocument.Source` property is a file on the local file system. I.e., this module should only be used after a ReadFiles module or some other module that reads file system files. Using it on documents that come from other sources like URLs might produce unexpected results.
- We'll also have to document that this essentially takes the metadata from documents already in the pipeline. I was initially confused and thought this module was reading from disk. I actually like your approach better because it means I can use whatever parsing method I want to generate the folder metadata (YAML, JSON, etc.).
- I see no reason this shouldn't just go right in Wyam.Core since it doesn't use any external dependencies. Go ahead and continue working in the separate projects and I'll stick them into Wyam.Core right after the PR gets merged.

Then the config file would look something like this:

``` C#
Pipelines.Add("Content",
  ReadFiles("*.*"),
  FrontMatter(Yaml()),
  Markdown(),
  DirectoryMeta()
    .WithMetadataFile("my-meta.yaml", override: true)
    .WithMetadataFile("my-inherited-meta.yaml", true)
  WriteFiles(".html")
);
```

LokiMidgard:
I relay like your api design. I will start changing the project.

I will also add what you pointed out to the source documentation.

LokiMidgard:
currently the strategy for conflicts is original metadata is preserved, then local directory metadata is used and after that the inhered metadata (using the one that is nearest).
With the new design this wouldn't work that nicely.

A new approach could be
- The module warns (error?) if two metadata files override the same key
- Define that the order in which the metadata file names were added will resolve this.
  - override does not affect it
  - metadata with override have the higher priority.
- Do both

Currently I prefer the second. With the first sub node.

LokiMidgard:
So 
- changed the interface
- Tests still green
- changed documentation

it would be nice if you could look at the documentation, I'm not sure if what I've written is understandable.  (¬_ ¬ ;)

LokiMidgard:
Just thought about the abstraction from file system and different things came in my mind.

## Does more Modules need a file system?

The GitContributor Module is interpreting the Source Property as path to a file.

I'm not a great fan of "by convention" or we don't know what happens. If there are more modules that need a specific interpretation of the Source Property, maybe we can add information what kind of path source is. There could be an enum linke this that will be used in a SorceKind Property of Document.

``` C#
[Flags]
enum SorceKind
{
    Custom = 0,
    Structured = 1,
    Absolute = 2,
    Relative = 4,
    Filesystem = 9,
    Url = 17,
    Namespaces = 33,
    // etc...
}
```

A Module that actually read files could check if the source is an actual file. Warning or ignoring if non file based documents will be processed. 

If we would consider this we should discuss what consequences this have in a separate issue.

## Does we need this kind of dependency that Source is a path of the file system?

Instead of using Path.Combine and Path.GetDirectory we could write our implementation that uses custom separator. I think the only requirement we actually need is that Sorce is somewhat structured. An Element can have a Parent and multiple child's.
The MetadataFile (rename to MetadataElement?) gives it Metadata to all other child's of his parent (silblings?).

daveaglick:
I think you raise some great questions. I also hate convention-based code, so I think this could certainly be improved. I've opened a cooresponding discussion issue at #88

daveaglick:
Thanks again - this one is merged. I moved the module to core and also renamed `@override` to `replace` - I've found that novice developers sometimes don't know about escaping reserved keywords and since this was going to be exposed through the docs, I didn't want anyone getting confused.

PR #87, opened 2015-10-25
Added Fork Module

LokiMidgard:
Well maybe ForkMerge or BranchMerge would be a better name.

It simply makes the same as Concat, but instead of a new empty document you get all the input.
I use this Module when generating Files based on the current input and than merging these files back to the original stream of files.


daveaglick:
Made a couple small changes. Named the module `BranchConcat` since it basically does what those two combined do. Also added a `Where()` condition and some tests.

PR #89, opened 2015-10-25
Added new Modules and Information to Git Modules

LokiMidgard:
## Added Modules
- GitCommits
  Creates a Document for every Commit made, adding the Metadata
  for the corespondending SHA and an array of CommitInformations.
- GitFileCommits
  Looks up what file is associated with each Document, similar to GitContributor
  but instead stores the complete CommitInformation not one per Author.
## Modified
- Added more information to CommitInformation
  - The sha of the commit
  - The sha's of the parents
- The GitContributer Module returned for every Author one CommitInformation
  (Which includes the Author). Previously it was not defined which CommitInformation
  was returnd. Now it is the last Commit Information.
- Two CommitInformations was reagerded to have the same Author if the name
  of the Author Property was the same. Now it uses the Email address.

### DIFF
--- a/Wyam.Modules.Git/CommitInformation.cs
+++ b/Wyam.Modules.Git/CommitInformation.cs
@@ -9,19 +9,25 @@ namespace Wyam.Modules.Git
 {
     public class CommitInformation
     {
+        private string Sha { get; }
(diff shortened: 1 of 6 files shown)


daveaglick:
Thanks as always!

PR #98, opened 2015-10-28
Fixed a bug and changed the interface a little bit.

LokiMidgard:
## Bug

It was the same Bug that existed in an other class, But this time in GitAuthor Module. I compared the Name of Author instead of email. Now Email will be used instead.
## Change of Class Author

Previously an `IEqualityComparer<Author>` was used to compare the email address instead of the default equals that was inhered from object. The reason was Author also head a field DateTime. Ignoring the DateTime on equality checks seems to be a bead idea.

To change this, the DteTime filed is now owned by CommitInformation. Allowing Author to override the equals Method in a way someone would expect.
### Future Changes that will have to be done

This Project defenetltily need some unit tests. I will check out if I can create some with some test repository that will be copied when testing. This could take some time.

### DIFF
--- a/Wyam.Modules.Git/Author.cs
+++ b/Wyam.Modules.Git/Author.cs
@@ -7,13 +7,35 @@ namespace Wyam.Modules.Git
     {
         public string Name { get; }
         public string Email { get; }
-        public DateTimeOffset DateTime { get; }
+
 
         internal Author(Signature committer)
         {
             Name = committer.Name;
             Email = committer.Email;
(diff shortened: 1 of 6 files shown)


daveaglick:
Re: unit testing - you've probably already seen this, but LibGit2Sharp looks like it actually has a pretty good set of unit tests: https://github.com/libgit2/libgit2sharp/tree/vNext/LibGit2Sharp.Tests Their testing techniques might work here as well.

PR #101, opened 2015-10-29
Added ThenBy for OrderBy Module

LokiMidgard:
- I added A Method ThenBy to the Order By Module.
- I changed Descending so it now sets Descending of the last ThenBy that was added if any, else it has the old meaning.

```
    OrderBy(@doc["m1"])
        .ThenBy(@doc["m2")
        .Descending()
```

Orders the documents first for m1 in ascending then for m2 in descending order.

```
    OrderBy(@doc["m1"])
        .Descending()
        .ThenBy(@doc["m2")
```

Orders the documents first for m1 in descending then for m2 in ascending order.


daveaglick:
This is an excellent addition, thanks!

PR #103, opened 2015-10-30
FileName module: added Hindi/Unicode test, added overload to the cons…

yetanotherchris:
FileName module: added Hindi/Unicode test, added overload to the constructor to allow certain characters to be allowed, e.g. "-" support for creating tag pages (as they're common in tags)

### DIFF
--- a/Wyam.Core/Modules/FileName.cs
+++ b/Wyam.Core/Modules/FileName.cs
@@ -17,7 +17,7 @@ namespace Wyam.Core.Modules
 {
     public class FileName : IModule
     {
-	    private readonly IEnumerable<string> _allowedCharacters;
+	    private IEnumerable<string> _allowedCharacters;
 
 	    internal static readonly string[] ReservedChars = new string[]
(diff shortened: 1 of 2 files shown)


Wyam.Core/Modules/FileName.cs
```diff
@@ -83,8 +85,13 @@ public FileName(string inputKey, string outputKey)
             _fileName = (d, c) => d.String(inputKey);
             _outputKey = outputKey;
         }
-        
-        public FileName(DocumentConfig fileName, string outputKey)
+
+		public FileName(IEnumerable<string> allowedCharacters) : this()
```
daveaglick: Good idea to add a whitelist - can we make this a fluent method instead: `WithAllowedChars()`? That way you can specify allowed characters regardless of which keys you specify to get or set in the existing constructors.

yetanotherchris:
Yeh that sounds a lot better, the constructor way did feel messy

PR #107, opened 2015-11-01
FileName module: Filenames now lowercased. Fixed WithAllowedCharacters so it…

yetanotherchris:
FileName module:
- Filenames now lowercased. 
- Fixed FileName.WithAllowedCharacters so it returns the current instance.


daveaglick:
Thanks!

PR #110, opened 2015-11-05
Added Sitemap-module plus Example project. Also added a new metadata …

reliak:
…key that gives the relative location of the destination file


Wyam.Core/Modules/WriteFiles.cs
```diff
@@ -155,6 +155,7 @@ public virtual IEnumerable<IDocument> Execute(IReadOnlyList<IDocument> inputs, I
                             Metadata.Create(MetadataKeys.DestinationFileName, Path.GetFileName(path)),
                             Metadata.Create(MetadataKeys.DestinationFileDir, Path.GetDirectoryName(path)),
                             Metadata.Create(MetadataKeys.DestinationFilePath, path),
+                            Metadata.Create(MetadataKeys.RelativeDestinationFilePath, path.Substring(context.OutputFolder.Length)),
```
daveaglick, line 158: Good idea - it makes total sense that you may need to know the relative path of a written file. ReadFiles already sets metadata for `RelativeFilePath`, `RelativeFilePathBase`, and `RelativeFileDir`. Any reason why we shouldn't just set the same metadata for consistency in WriteFiles?

daveaglick:
This looks fantastic, thanks! And especially thanks for the example - that should get folks up and running with it very quickly.

A few comments:
- I added a comment in the changeset about the metadata for WriteFiles - maybe we can reuse the same Relative... metadata keys as ReadFiles?
- Does this module use any external libraries (it doesn't appear so)? If not, then I'll probably move it into Wyam.Core after merging (but you can keep it separate until then).
- Good work on providing the location formatter to ensure the `loc` values can contain absolute URLs.
- Since `loc` is the only required sitemap element, can we provide an alternate constructor that doesn't require specifying a `SitemapItem` metadata key and instead just passes each document itself to the location formatter? Don't get me wrong, I like the way this was set up with a special object to store in metadata with the needed data, but if all we care about is `loc` then creating and storing a `SitemapItem` might be overkill.

daveaglick:
Even better, leave the constructor as-is and then in the `Execute()` method when you get the corresponding "SitemapItem" metadata value, do one of three things:
- If the metadata contains an actual `SitemapItem`, process as you currently do.
- If the metadata doesn't contain a `SitemapItem`, send the `.ToString()` value of whatever it does contain to the location formatter.
- If the "SitemapItem" key doesn't exist or is null, skip the document entirely.

Wyam.Core/Modules/WriteFiles.cs
reliak, line 158: I think it's ok if `WriteFiles` overwrites the metadata for `RelativeFilePath` etc. If there would be a usecase where the original pathes are needed _after_ `WriteFiles`, they still can be copied before `WriteFiles` to metadata under a different name.

reliak:
Thanks for your feedback.
The module has no external dependencies, so it could be safely used within core.

I share your opinion that the creation of a `SitemapItem` might be overkill for situations where only the location is needed.
Your suggested approach seems reasonable. I have a slight modification to it.
First, as a general suggestion, I think it would be nice to encourage users to specify the hostname of the target website as metadata in the `wyam.config` (`Metadata["Hostname"] = "http://www.example.com";`).
This hostname can then be used throughout all the modules, including the sitemap module.

Second, the sitemap module could check, whether a given sitemap `Location` is an absolute or relative url. For the latter it could combine the hostname from the metadata with the relative url.
The location formatter would then only be needed for special cases.

daveaglick:
That's a great idea - I wouldn't want to force it, but we can certainly encourage that builds intended as public web sites use it. Once it's documented as the "official" way to specify a host name, modules could then feel comfortable checking for it (though will obviously still need a fallback if it doesn't exist).

The only complication I can see is the local testing server - if modules get too comfortable with the idea of using host name, then local testing might break. For example, it's not a problem for the site map since that's not really useful on your local system, but what if some module that generates links prepends the host name to them instead of making them relative. I'll just be careful in the docs to say that modules should only use the host name to create an absolute path if it's actually necessary.

PR #111, opened 2015-11-06
Updating RelativeFilePath after WriteFiles

reliak:


daveaglick:
FYI - I went ahead and added the other Realtive\* metadata values to WriteFiles. I also slightly modified the behavior of RelativeFilePath to ensure consistency between ReadFiles and WriteFiles - that way downstream modules will get the same value for the relative path metadata and don't need to care whether the document has gone through a WriteFiles module.

daveaglick:
BTW, thanks!

reliak:
Thanks for merging.

I do have another suggestion. Is there a reason to use the backslash ('\') instead of the slash ('/') for files and directories?
I mean, .NET IO can handle normal slashes too (e.g. File.WriteAllText("c:/subfolder/test.txt") works fine).
The benefit would be, that the relative path is also usable as a url. Currently I need to call PathHelper to replace backslashed with slashes.

daveaglick:
No reason at all other than that I struggle with the "right" way to do file I/O. It think the only requirement should be that users can input/specify paths with either type of slash (which should then get normalized to some canonical style).

As long as it's documented and consistent, I don't think it matters which direction they're normalized to. You make an excellent point about the benefits of normalizing to a URL-style slash (I've had trouble with that in the past too). Care to take a look and see what it would take to change all uses of backslash to slash?

This is also a good thing to take another look at because I've had it in the back of my mind that Wyam will eventually end up supporting xplat .NET. Probably another reason to standardize on a normal slash.

reliak:
> Care to take a look and see what it would take to change all uses of backslash to slash?

I looked through the projects and I didn't see any problematic backslashes in the productive code. There are some backslashes used in test code, but I don't see that as critical.

PR #113, opened 2015-11-07
Added updated Sitemap to the Wyam core

reliak:
I created a new branch on my fork for this and closed the old pull request (https://github.com/Wyamio/Wyam/pull/110) for the sitemap feature.

Looking forward for your comments on this.


daveaglick:
:thumbsup: - so close! I've only got two more small comments I'll make online with the code. Normally I wouldn't be looking so closely, but I think this module is going to become one of the "standard ones" that just about everyone uses, so I want to make sure it tackles the broadest number of use cases.

reliak:
Looking closely is good. You know the infrastructure of the project better than me and know whats best.

Wyam.Core/Modules/Sitemap.cs
```diff
@@ -0,0 +1,106 @@
...
+        public Sitemap(string sitemapFilename, Func<string, string> locationFormatter = null)
+        {
+            this.sitemapFilename = sitemapFilename;
+            this.locationFormatter = locationFormatter;
+        }
+
+        public IEnumerable<IDocument> Execute(IReadOnlyList<IDocument> inputs, IExecutionContext context)
+        {
+            var outputFilename = Path.Combine(context.OutputFolder, this.sitemapFilename);
+            var docs = context.Documents.Where(f => f.Metadata.ContainsKey(MetadataKeys.SitemapItem)).ToArray();
```
daveaglick: Can we use `inputs` here instead of getting all documents from the context? That'll also require a `Documents()` module in the example pipeline to feed the site map.

This will allow fine-grained control of which documents the site map covers. It'll also work better for scenarios with "throw away" documents (for example, in the new Wyam web site I'll be launching soon, one pipeline finishes with source code documents that are only meant to drive other pipelines, not be output as pages).

Wyam.Core/Modules/Sitemap.cs
```diff
@@ -0,0 +1,106 @@
...
+                foreach (var doc in docs)
+                {
+                    object hostname = null;
+                    doc.Metadata.TryGetValue(MetadataKeys.Hostname, out hostname);
+
+                    var smi = doc.Metadata.Get(MetadataKeys.SitemapItem);
+                    var itm = smi as SitemapItem;
+
+                    if (itm == null)
+                        itm = new SitemapItem(smi.ToString());
```
daveaglick, line 54: Can we also handle the case where `smi` is null by using the value for `RelativeFilePath` (assuming it exists)? That way the user doesn't even need to make a `SitemapItem` metadata key if they don't want to.

Wyam.Core/Modules/Sitemap.cs
```diff
@@ -0,0 +1,106 @@
...
+                    var location = itm.Location;
+
+                    if (this.locationFormatter != null)
+                        location = this.locationFormatter(location);
+                    else
+                    {
+                        if (hostname != null && !location.StartsWith("http://", StringComparison.InvariantCultureIgnoreCase))
+                            location = $"{hostname}{PathHelper.ToRootLink(location)}";
+                    }
+
```
daveaglick, line 73: Should probably have a `if(location == null)` check here and skip the rest of the site map item if the delegate returned null (which would be another way to control which documents appeared in the site map).

Wyam.Core/Modules/Sitemap.cs
reliak, line 54: That would only work when using your suggested inputs method using `Documents` before `Sitemap`. Otherwise all documents from all pipelines that have a value for `RelativeFilePath` would be added to the sitemap.

Allowing to omit the `SitemapItem` meta data would make it impossible to exclude some urls from a pipeline. In my example project I used the `If` module to omit the `SitemapItem` meta data from specific urls. This wouldn't be possible when we would use only `RelativeFilePath`. Do you agree on this?

Wyam.Core/Modules/Sitemap.cs
reliak, line 73: Yeah, I can do that.

Wyam.Core/Modules/Sitemap.cs
daveaglick, line 54: Yeah, that makes sense. Go ahead and leave it out for now considering our conversation on Gitter.

Issue #93, opened 2015-10-27
ReadXmp Module


LokiMidgard:
A Module that Extract [XMP](https://en.wikipedia.org/wiki/Extensible_Metadata_Platform) Information from files. This could be very useful to automatically show the Author and License for a picture or other media.

It could probably based on [metadata-extractor](https://github.com/drewnoakes/metadata-extractor-dotnet).

daveaglick:
Nice work!

PR #114, opened 2015-11-07
Xmp Module

LokiMidgard:
Created a Xmp Module #93 

You can specify what xmp tags you are interested in and what name the key in the document metadata should have.

This Module supports reading embedded metadata from supported files like .png and .jpg and sidecarfiles with the .xmp extension.

```
Xmp(errorsOnDoubleKeys:false)
    .WithMetadata("cc:license", "Copyright")
    .WithMetadata("xmpRights:UsageTerms", "Copyright")
    .WithMetadata("dc:rights", "Copyright")
    .WithMetadata("dc:creator", "Creator")
    .WithMetadata("cc:attributionName", "Creator")
    .WithMetadata("xmpRights:Marked", "RightsReserved"),
```

This sample reads data from `dc:rights`, `xmpRights:UsageTerms` and `cc:license` to the document metadata with the key `"Copyright"`. In this sample the later the Metadata is defined the higher its priority. So `dc:rights` overrides `cc:license`

You need to specify the namespaces for the xmp tags. The Module supports some default like dc and cc. But you can define and override any namespace.


daveaglick:
:thumbsup: thanks! I've got to head out in a minute, but hopefully I'll have a chance to review/merge later today.

daveaglick:
It looks like there's two Xmp modules, one in Wyam.Modules.Images and one in Wyam.Modules.Xmp. I suspect the one in Wyam.Modules.Images is old/wrong (it looks incomplete)?

Wyam.Modules.Xmp/Xmp.cs
```diff
@@ -0,0 +1,353 @@
...
+            _skipElementOnMissingData = skipElementOnMissingMandatoryData;
+            _errorOnDoubleKeys = errorsOnDoubleKeys;
+            _delocalizing = delocalizing;
+            _flatting = flatten;
+
+            namespaceAlias["dc"] = "http://purl.org/dc/elements/1.1/";
+            namespaceAlias["xmpRights"] = "http://ns.adobe.com/xap/1.0/rights/";
+            namespaceAlias["cc"] = "http://creativecommons.org/ns#";
+            namespaceAlias["xmp"] = "http://ns.adobe.com/xap/1.0/";
+            namespaceAlias["xml"] = "http://www.w3.org/XML/1998/namespace";
```
daveaglick: If these are all always used as initial values, can we move them up to the declaration of `namespaceAlias` as initializer syntax to declutter the constructor? I.e.,:

``` C#
private readonly Dictionary<string, string> namespaceAlias =
    new Dictionary<string, string>()
    {
        {"dc", "http://purl.org/dc/elements/1.1/"},
        {"xmpRights", "http://ns.adobe.com/xap/1.0/rights/"},
        // Etc.
    };
```

Wyam.Modules.Xmp/Xmp.cs
```diff
@@ -0,0 +1,353 @@
...
+                         context.Trace.Warning($"File doe not contain Metadata or sidecar file ({x.Source})");
+                         if (_skipElementOnMissingData)
+                             return null;
+                     }
+                     return x;
+                 }
+
+                 Dictionary<string, object> newValues = new Dictionary<string, object>();
+
+                 TreeDirectory hirachciDirectory = TreeDirectory.GetHirachicDirectory(xmpDirectory);
```
daveaglick: `hierarchicalDirectory` and `GetHierarchicalDirectory()` :)

daveaglick:
Now that I've had a chance to look over the whole thing, this is excellent work. And it looks like it probably wasn't easy to get all the aspects of XMP to work correctly. Thanks for sticking with it, it'll be a valuable addition.

Wyam.Modules.Xmp/Xmp.cs
LokiMidgard: That's seems to be a good idea, it looks much nicer.

LokiMidgard:
Your right I didn't notice the second Xmp Module under image. I will remove it.
Also changing the tow things you pointed out. Thank you for the valuable advices.

LokiMidgard:
So fixed the parts you mentioned.

PR #116, opened 2015-11-08
Updated CopyFiles module to handle specified subfolders

reliak:
(e.g. CopyFiles("subfolder/_._"))

Also added CopyFilesFixture.cs to demonstrate this use case


Wyam.Core/Modules/CopyFiles.cs
```diff
@@ -106,7 +106,7 @@ public IEnumerable<IDocument> Execute(IReadOnlyList<IDocument> inputs, IExecutio
                             .Select(file =>
                             {
                                 string destination = _destinationPath == null
-                                    ? Path.Combine(context.OutputFolder, PathHelper.GetRelativePath(Path.GetDirectoryName(path), Path.GetDirectoryName(file)), Path.GetFileName(file)) 
+                                    ? Path.Combine(context.OutputFolder, PathHelper.GetRelativePath(context.InputFolder, Path.GetDirectoryName(file)), Path.GetFileName(file))
```
daveaglick: What happens if the source path for the files is outside the `InputFolder`? Would this still work? If not, should we do a check to see if the source path is under `InputFolder` and do the new behavior if so and the old behavior if not?

reliak: You're right. The code assumes that the file to copy comes from the input folder.
If the user specifies a relative path we can assume that this should point to the input folder, right?
If so, we would only need to check whether the path is absolute or not.
This wouldn't work if the user is allowed to define relative paths like `../../` to move up the directory hierarchy.

daveaglick: I think checking for relative path is fine. Allowing ascention up the directory tree seems like such an edge case, I don't think we need to support it - at least not right now.

daveaglick: Actually, on second thought, maybe we should support that if we can. I can see a use case where you bring in a folder alongside the Input and Output folders at the root of the project via git submodule or something and then want to copy over stuff inside it.

Would doing a `StartsWith()` to compare the input path with the copy source work for the test?

reliak: I think we first need a method to detect and convert special cases like `../../` to an absolute path.
After that, it's enough to check for relative paths.

daveaglick: That makes sense, especially since we'll have to convert it eventually any way.

In fact, one thing I want to look at soon is a better abstraction for the file system to take the place of System.IO usages. I especially like the one used in Cake, so may take some inspiration there.

reliak: Ok, I will update the PR accordingly.

> In fact, one thing I want to look at soon is a better abstraction for the file system to take the place of System.IO usages. I especially like the one used in Cake, so may take some inspiration there.

That would be nice. That should also make unit testing easier.
As I saw, Cake uses static classes for the abstraction. I would prefer a more interface-based approach. Have you thought about bringing dependency injection into Wyam?

daveaglick: Yeah, whatever we end up with will definitely be interface based, with the interfaces going in Wyam.Common and the implementations going in Wyam.Core, similar to the rest of the abstractions. Reliance on static members is one reason I want to get off of System.IO in the first place. I envision something like a `IFileSystem` object being passing in the execution context as the entry point to file system functionality.

Gave DI a lot of thought when first architecting things. In general, I'm not a fan of canned DI frameworks (though I think the overall concept of IoC and runtime implementation selection is fine). Usually too much "magic" for my taste. You can see I like to be very explicit - heck, I'm still not a big fan of `var` :)

That said, there are two places I could see wanting to swap out different implementations for a few select interfaces: testing and the config files (for really advanced scenarios). What I may end up implementing at some point soon is a more direct factory pattern with implementation registration in the engine.

reliak: Yeah, an abstraction as in `IFileSystem` sounds good.

I updated the PR btw.

daveaglick:
Great work, thanks! And once again, thanks for including tests.

PR #120, opened 2015-11-09
Added Switch module to support switch-statement logic for modules

reliak:
This module should make cases easier, where a `If` module would be too verbose, i.e. to reduce redundant code.
An example would be to use the `Switch` module, to do certain operations depending on the file extension.
Such an example is shown below:

```
// assuming we have four input files: test.txt, test.css, test.xml and test.js,
// which all contain the text "test"
Pipelines.Add("SwitchDemo",
    ReadFiles("*.*"),
    Switch( ((string)@doc["SourceFileExt"]).ToLower() )
    .Case(".js", ReplaceIn("test", "js"))
    .Case(new []{".txt", ".css"}, ReplaceIn("test", "txt or css"))
    .Default(ReplaceIn("test", "default for all other extensions")),
    WriteFiles()
);
```

The fluent `Default` method is optional. Without `Default` method, any document that is not catched by the defined cases, will not be modified.

### DIFF
--- a/Wyam.Core.Tests/Modules/SwitchFixture.cs
+++ b/Wyam.Core.Tests/Modules/SwitchFixture.cs
@@ -1,0 +1,124 @@
+﻿using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


daveaglick:
This is excellent, thanks.

PR #125, opened 2015-11-14
Documentation

LokiMidgard:
Added Documentation to Git and XMP modules.


daveaglick:
This was super helpful, thanks. Getting all the doc content into XML comments so the site can keep up with changes is the last little thing before release.

PR #126, opened 2015-11-19
Xslt module

LokiMidgard:
Created Module to transform xml using xlst.


daveaglick:
This was a great idea - I did a little XSLT back in the day, and I know a lot of older stuff is still heavily powered by it.

daveaglick:
Since there's no external dependencies, I may move the module into Wyam.Core. Shouldn't affect any config files you've already written.

LokiMidgard:
Your right it only uses standard .Net stuff.

PR #132, opened 2015-11-20
Created two module excel to csv and csv to html table

LokiMidgard:
Created two Modules that will take a document and translate it from xslx to csv and from csv to a html table tag.


daveaglick:
:thumbsup:

PR #133, opened 2015-11-21
Added some comments to the (Metadata)Keys class.

LokiMidgard:
I added some doc comments to the Keys class.

I'm not sure if they are correct.
-  I didn't now the difference between `SourceFileRoot` and `SourceFileDir`.
- Are relative path based on destination or source keys (I assumed source).
- What is the difference between `destination-` and `write-` metadata.

### DIFF
--- a/Wyam.Common/Meta/Keys.cs
+++ b/Wyam.Common/Meta/Keys.cs
@@ -18,15 +18,45 @@ namespace Wyam.Common.Meta
         public const string Hostname = "Hostname";
 
         // ReadFile/WriteFiles/CopyFiles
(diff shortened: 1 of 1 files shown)


daveaglick:
The best place to look for information about metadata is on the modules that use them with the new custom `<metadata ...>` XML doc comment (I.e., https://github.com/Wyamio/Wyam/blob/develop/Wyam.Core/Modules/IO/ReadFiles.cs#L26). The docs site knows how to read these so that's what'll show up on the Wyam web site for each module.

LokiMidgard:
Cool :)

Issue #147, opened 2015-12-18
Engine.Configure() method without parameters don't add any namespace into ExecutionContext


heavenwing:
This issue also lead to any Razor View cannot be compiled.

daveaglick:
Fixed by #148

PR #148, opened 2015-12-18
fix issue #147

heavenwing:
Engine.Configure() method without parameters don't add any namespace into ExecutionContext


daveaglick:
Great catch, thanks! Everything looks good - I'll merge it in tomorrow. I should be able to take a closer look at your other issue then too.

Issue #27, opened 2015-06-16
Create an RSS feed module


daveaglick:
Constructs an RSS feed from the input documents. How to handle feed items from multiple pipelines (might not need to worry about it if there's also a module that can "reuse" documents from other pipelines)? How to extract the excerpt (pull from metadata or provide a `Func<IDocument, string>`)?

daveaglick:
Note that `System.ServiceModel.Syndication` isn't yet available in .NET Core (and it's unclear when it will be).

miere43:
I guess this one should be closed because this module already present #174

daveaglick:
Good call, thanks again!

PR #174, opened 2015-12-30
New RSS module

miere43:
Module for generating RSS and Atom feeds from input documents. Not all features implemented, just basics. Output RSS feeds are passing [W3 RSS validator](https://validator.w3.org/feed/check.cgi). #27

### DIFF
--- a/Examples/Rss/Input/posts/flex.md
+++ b/Examples/Rss/Input/posts/flex.md
@@ -1,0 +1,7 @@
+---
+Title: New Year must be awesome
+Description: (surely)
+Published: 01/01/16
+---
+
+Wat ar u talkin' about?
--- a/Examples/Rss/Input/posts/hello.md
+++ b/Examples/Rss/Input/posts/hello.md
@@ -1,0 +1,6 @@
+---
+Title: Hello, world!
+Published: 02/01/16
+---
+
+Hello, world!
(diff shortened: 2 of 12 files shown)


daveaglick:
This looks really excellent, thanks! I especially appreciate the example and the detailed documentation. I'll merge this as soon as I'm at my desk.

Out of curiosity, was there a reason you chose not to use the classes from `System.ServiceModel.Syndication` (I.e., something like https://github.com/daveaglick/daveaglick/blob/master/Input/rss.cshtml)? Did you get more control by constructing the feed directly?

miere43:
@daveaglick, no advantages in writing it directly, just I had more experience with plain XML construction rather than using Syndication classes. Through I think it's better to rewrite with Syndication, because many methods are available out of box, [like formatting DateTime](https://github.com/miere43/Wyam/blob/develop/Wyam.Modules.Rss/DateTimeRssExtensions.cs).

daveaglick:
I'm going to merge this as-is for now and we can reconsider a switch to the syndication libraries later. I'd like to port over to .NET Core at some point next year, and those classes aren't available yet (see https://github.com/dotnet/wcf/issues/76). Not sure if or when they'll make it over, so the safe bet is probably to stick with raw XML generation until we know more.

PR #203, opened 2016-02-02
Added a IsRegex option to Replace.

deanebarker:
As discussed on Gitter.


daveaglick:
This will come in very handy, thanks!

PR #220, opened 2016-02-12
Added a ValidateMeta module.

deanebarker:


deanebarker:
I commented out the test.  It was failing, but it's because I don't understand NSubstitute, not because the logic is wrong.

daveaglick:
This was a great contribution, thanks! Another one I'm looking forward to integrating in my own projects. I'll take a look at the test (and maybe add a couple more).

PR #223, opened 2016-02-12
Added a "--verify-config" command line switch.

deanebarker:
In the end, all this does is exit before execution starts.  Nothing else is changed.


daveaglick:
:+1:

PR #232, opened 2016-02-16
Highlight exceptions stack trace in verbose mode

### LINKED ISSUE #230: CLI: Stack trace doesn't being highlighted when verbose mode turned on
See `Exception: System.AggregateException: ...`. This string should be in red.

miere43:
Resolves #230


daveaglick:
Great solution - I was wondering how I was going to change the colors for a single trace message, this handles it nicely :+1:

PR #237, opened 2016-02-17
Ability to pass CLI args to config

miere43:
Currently works like that: `wyam.exe --pass define --pass key=value --pass Key=Value`
- key/values are case-sensitive
- if you pass just a key, value will be null
- can be accessed from config.wyam using `EngineMetadata` variable (type `Dictionary<string, string>`)
- you **can't** pass like that: `wyam.exe --pass "key=value,key2=value2"`
- no shortcut for command line command
- you can escape `=` key by writing backslash before it: `this\=is_key` => `KeyValuePair<string, string>("this=is_key", null)`;
### Some thoughts
- you still can't get other command like args like `--watch` or `--preview`
- you don't know is metadata truly passed because if metadata key is invalid this is not passed to `EngineMetadata` and just being dropped. You can see a warning in CLI, but if you are executing Wyam from elsewhere (from another program) you can't get a warning. **Solution**: add a switch to make Wyam stop executing if key was dropped and create new return code.

### DIFF
--- a/src/Wyam.Core.Tests/TestGlobalMetadataParser.cs
+++ b/src/Wyam.Core.Tests/TestGlobalMetadataParser.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using NUnit.Framework;
+using Wyam.Core.Util;
+
+namespace Wyam.Core.Tests
+{
+   public class TestGlobalMetadataParser
+    {
+        GlobalMetadataParser parser;
+
+        [SetUp]
+        public void SetUp()
+        {
+            parser = new GlobalMetadataParser();
+        }
+
+        [Test]
+        public void TestKeyOnlyParse()
+        {
+            var excepted = new string[] { "hi", "=hello", "\\=abcd", "key\\=val", "     bjorn  \\=   dad" };
+
+            var args = parser.Parse(excepted);
+            Assert.AreEqual(excepted.Length, args.Count);
+
+            int i = 0;
+            foreach (var arg in args) {
+                Assert.AreEqual(excepted[i].Replace("\\=", "=").Trim(), arg.Key);
+                Assert.IsNull(arg.Value);
+                i++;
+            }
+        }
+
+        [Test]
+        public void TestKeyValueParse()
+        {
+            var pairs = new string[] { "key=value", "k=v", "except=bro", "awesome====123123", "   keytrimmed    =    value trimmed   " };
+            var args = parser.Parse(pairs);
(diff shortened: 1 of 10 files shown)


daveaglick:
This is awesome! :+1:

One small request though - I'm not crazy about the `--pass` argument name. Would you object to calling it `--meta`? I think that would make it clearer what's going on.

Also wondering if we should just add these to the `InitialMetadata` collection instead of a new one. That way they're available in the config file, and also from `IExecutionContext` inside modules. Thinking is that it shouldn't matter whether metadata is coming from the command line, from the config file, or from somewhere else - it all ends up in the same place (and gets added to each new document).

miere43:
I'll change the argument name.

I though about same `InitialMetadata` thing, but I saw it is being passed to `DocumentFactory` and then from it to `Document`. I think CLI passed metadata shouldn't belong to document, so I leaved `InitialMetadata` field as is.

With this implementation a few questions left:
- Should Wyam stop execution when invalid meta passed instead of throwing a warning?
- `EngineMetadata` should be of type `Dictionary<string, string>` or `IMetadata` instead? I didn't clearly got purpose of `IMetadata`, so I leaved a dictionary. It would be nice to switch to `IMetadata` because it contains really neat methods to convert values.

daveaglick:
> I think CLI passed metadata shouldn't belong to document

I can envision use cases where it would be helpful, and I can see some where it would actually be bad. Perhaps the best approach is to leave it as you wrote it and then consider how to satisfy the case where you want metadata do flow from the CLI to the documents in a separate issue. I could see either adding another flag or engine method that essentially copies metadata from the `EngineMetadata` to the `InitialMetadata` or adding an extra CLI option that does initial metadata instead of engine metadata. Let's not worry about that for now...

> Should Wyam stop execution when invalid meta passed

My gut says yes. Wyam used to be more forgiving of errors, but that doesn't make a lot of sense for a generator. If something is wrong, that means there's unexpected behavior. Especially if it's something like data you're supplying over the command line. If there's something off with it, I think we've got to kill the build.

> `EngineMetadata` should be of type `Dictionary<string, string>` or IMetadata instead?

Agree there is a lot of cool conversion report in the built-in metadata. However, it's also a big surface area to implement in order to get that support. I would say just leave it as a dictionary for now. What's more important is that we can get the values out exactly as they were specified on the command line. After that, the configuration script (or embedded host) should operate on the string values as needed.

miere43:
Seems okay now.

PR #243, opened 2016-02-22
Fixed multiple console tracers for engine

miere43:
Fixed engine trace listener was not removed when requesting new engine which lead to double tracings in console while in --watch mode.

```
...
    Executing pipeline "Resources" (4/5) with 1 child module(s)
    Executing pipeline "Resources" (4/5) with 1 child module(s)
    Executing pipeline "Resources" (4/5) with 1 child module(s)
    Executing pipeline "Resources" (4/5) with 1 child module(s)
    Executing pipeline "Resources" (4/5) with 1 child module(s)
        Executed pipeline "Resources" (4/5) in 18 ms resulting in 38 output document(s)
        Executed pipeline "Resources" (4/5) in 18 ms resulting in 38 output document(s)
        Executed pipeline "Resources" (4/5) in 18 ms resulting in 38 output document(s)
        Executed pipeline "Resources" (4/5) in 18 ms resulting in 38 output document(s)
        Executed pipeline "Resources" (4/5) in 18 ms resulting in 38 output document(s)
    Executing pipeline "Resources (Less)" (5/5) with 3 child module(s)
    Executing pipeline "Resources (Less)" (5/5) with 3 child module(s)
    Executing pipeline "Resources (Less)" (5/5) with 3 child module(s)
    Executing pipeline "Resources (Less)" (5/5) with 3 child module(s)
    Executing pipeline "Resources (Less)" (5/5) with 3 child module(s)
        Executed pipeline "Resources (Less)" (5/5) in 18 ms resulting in 1 output document(s)
        Executed pipeline "Resources (Less)" (5/5) in 18 ms resulting in 1 output document(s)
        Executed pipeline "Resources (Less)" (5/5) in 18 ms resulting in 1 output document(s)
        Executed pipeline "Resources (Less)" (5/5) in 18 ms resulting in 1 output document(s)
        Executed pipeline "Resources (Less)" (5/5) in 18 ms resulting in 1 output document(s)
    Executed 5/5 pipelines in 420 ms
    Executed 5/5 pipelines in 420 ms
    Executed 5/5 pipelines in 420 ms
    Executed 5/5 pipelines in 420 ms
    Executed 5/5 pipelines in 420 ms
Hit any key to exit
Hit any key to exit
Hit any key to exit
Hit any key to exit
Hit any key to exit
```

### DIFF
--- a/src/Wyam/Program.cs
+++ b/src/Wyam/Program.cs
@@ -56,6 +56,7 @@ namespace Wyam
         private DirectoryPath _previewRoot = null;
         private FilePath _configFilePath = null;
         private IReadOnlyList<string> _globalRawMetadata = null;
+        private System.Diagnostics.ConsoleTraceListener _engineListener = null;
 
         private readonly ConcurrentQueue<string> _changedFiles = new ConcurrentQueue<string>();
         private readonly AutoResetEvent _messageEvent = new AutoResetEvent(false);
@@ -356,7 +357,10 @@ namespace Wyam
                 Engine engine = new Engine();
 
                 // Add a default trace listener
-                Trace.AddListener(new SimpleColorConsoleTraceListener() { TraceOutputOptions = System.Diagnostics.TraceOptions.None });
+                if (_engineListener != null)
+                    Trace.RemoveListener(_engineListener);
+                _engineListener = new SimpleColorConsoleTraceListener() { TraceOutputOptions = System.Diagnostics.TraceOptions.None };
+                Trace.AddListener(_engineListener);
 
                 // Set verbose tracing
                 if (_verbose)


daveaglick:
Another great catch! This is happening only when the engine gets re-instantiated, which itself only happens when the config file changes which is why I didn't catch it (changes to other files won't result in a new engine). The bug was introduced because `Trace` switched to a thread-safe static in the last version instead of the engine instance that it was. We actually no longer need to mess with adding and removing listeners at all - they can just be added once at the start.

There's a better way to resolve this that doesn't require adding and removing the listener, so I'm going to close the PR, but thanks a ton for pointing out the problem!

PR #247, opened 2016-03-04
Mark build.sh as executable

eatdrinksleepcode:
The build.sh file needs to be marked as executable so that it can be run by default on Unix shells.

This change was made by running "chmod u+x build.sh".


daveaglick:
:thumbsup:

I fear you may not get very far on OS X though. It's in the plans to port everything over to Core later this year (once Core stabilizes - I just don't have the energy to play the moving target game). Right now I haven't done any testing or running on non-Windows at all. The build.sh script is more of an artifact of the Cake build process than an indication it's intended for Linux use.

That said - who knows! I'd love to hear if it runs on Mono (assuming that's what you're looking at?)

eatdrinksleepcode:
Understood. I appreciate that you have the build.sh script even if it is not fully working; it's a lot easier than starting from scratch.

Issue #246, opened 2016-03-04
NullReferenceException running build.sh on OS X


eatdrinksleepcode:
This is probably a Cake issue, but I am filing it here to bring it to your attention and maybe get help filing it with the Cake team.

Environment: OS X El Capitan, Mono 4.1.0

Running ./build.sh -v diagnostic results in the following error:

```
Error: System.NullReferenceException: Object reference not set to an instance of an object
  at Cake.Scripting.Mono.CodeGen.Parsing.ScriptParser.CreateBlock (Cake.Scripting.Mono.CodeGen.Parsing.ScriptToken current, Boolean hasScope)<0x10f3c0fc0 + 0x0002a> in <filename unknown>:0
  at Cake.Scripting.Mono.CodeGen.Parsing.ScriptParser.CreateBlock (Boolean hasScope)<0x10f3c0f80 + 0x00028> in <filename unknown>:0
  at Cake.Scripting.Mono.CodeGen.Parsing.ScriptParser.ParseNext ()<0x10f3c06b0 + 0x0018b> in <filename unknown>:0
  at Cake.Scripting.Mono.CodeGen.MonoScriptProcessor.ParseBlocks (Cake.Core.Scripting.Script script)<0x10f3bfd00 + 0x000fe> in <filename unknown>:0
  at Cake.Scripting.Mono.CodeGen.MonoScriptProcessor.Process (Cake.Core.Scripting.Script script, IReadOnlyList`1& blocks)<0x10f3bfa30 + 0x000d4> in <filename unknown>:0
  at Cake.Scripting.Mono.CodeGen.MonoCodeGenerator.Generate (Cake.Core.Scripting.Script script)<0x10f3bf590 + 0x00049> in <filename unknown>:0
  at Cake.Scripting.Mono.MonoScriptSession.Execute (Cake.Core.Scripting.Script script)<0x10f3bf380 + 0x00054> in <filename unknown>:0
  at Cake.Core.Scripting.ScriptRunner.Run (IScriptHost host, Cake.Core.IO.FilePath scriptPath, IDictionary`2 arguments)<0x10e88db60 + 0x00c21> in <filename unknown>:0
  at Cake.Commands.BuildCommand.Execute (Cake.CakeOptions options)<0x10e88daa0 + 0x00046> in <filename unknown>:0
  at Cake.CakeApplication.Run (IEnumerable`1 args)<0x10e887c30 + 0x000b4> in <filename unknown>:0
```

[...]

PR #248, opened 2016-03-04
Fix for verbatim strings in build script.

patriksvensson:
Since the Mono compiler and Roslyn differ in how they treat scripts, we have to do some script massaging before compiling. In this case you use verbatim strings to escape some backslashes that upset our block parser.

For some reason escaping the backslash didn't work either (have to look into this some more) so I resorted to changing them to forward slashes that solved the problem. This should to my knowledge be no problem for the nuspec packing.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -196,7 +196,7 @@ Task("Create-Tools-NuGet-Package")
             { 
                 new NuSpecContent 
                 { 
-                    Source = @"bin\" + configuration + @"\**\*",
+                    Source = "bin/" + configuration + "/**/*",
                     Target = "tools"
                 } 
             }


patriksvensson:
Should resolve #246.

daveaglick:
Awesome, thanks. This would have taken me forever to figure out.

patriksvensson:
@daveaglick You should verify that the nuget package you build still looks like it's supposed to :smile_cat:

daveaglick:
Yolo

daveaglick:
@patriksvensson Turns out NuGet doesn't like the slashes in the other direction. Instead of adding all the content at the root of the tools folder inside the package, it now adds it inside tools, but relative to the base (which nests everything a couple levels). You've got me on the right path though, so I'll play around with a few alternatives to see if I can get it to package correctly on Mono.

patriksvensson:
@daveaglick I can take a look as well and see if we can solve it.

eatdrinksleepcode:
For what it's worth I spent some time investigating a similar issue in my own project. Ultimately I was not able to figure out any way to package on both Mac/Mono and Windows/.NET except to have different slashes for each. To accomplish that I made the entire path a variable that gets passed to Nuget from my build scripts. See:

[NuSpec](https://github.com/eatdrinksleepcode/casper/blob/052c7c9dced7d01a133f54870675e01f4eeaee8e/Console/Console.nuspec#L15)
[pack (for Linux)](https://github.com/eatdrinksleepcode/casper/blob/052c7c9dced7d01a133f54870675e01f4eeaee8e/pack#L8)
[pack.cmd (for Windows)](https://github.com/eatdrinksleepcode/casper/blob/052c7c9dced7d01a133f54870675e01f4eeaee8e/pack.cmd#L9)

PR #249, opened 2016-03-04
Another fix for the backslash/verbatim problem.

patriksvensson:
Ok, this seems to solve it.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -185,7 +185,8 @@ Task("Create-Tools-NuGet-Package")
         if(nuspec == null)
         {            
             throw new InvalidOperationException("Could not find tools nuspec.");
-        }            
+        }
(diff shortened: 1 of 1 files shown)


daveaglick:
Much better! Although I kind of like the ugly hack I came up with too (which also looks like it works under "-mono"):

```
string.Concat("bin", (char)92, "Release", (char)92, "**", (char)92, "*")
```

:smile:

patriksvensson:
Haha, very creative!

I can't wait for the day when they announce xplat Roslyn (the scripting bits) so we don't have to deal with those kind of problems anymore :)

daveaglick:
Confirmed working as expected under both script engines. Thanks again for the help!

PR #269, opened 2016-04-06
Fixed bug when engine metadata was not retained after engine reset in…

miere43:
Fixed bug when engine metadata was not retained after engine reset in watch mode.

### DIFF
--- a/src/Wyam/Program.cs
+++ b/src/Wyam/Program.cs
@@ -55,6 +55,7 @@ namespace Wyam
         private DirectoryPath _previewRoot = null;
         private FilePath _configFilePath = null;
         private IReadOnlyList<string> _globalRawMetadata = null;
+        private Common.Meta.ISimpleMetadata _engineGlobalMetadata = null;
 
         private readonly ConcurrentQueue<string> _changedFiles = new ConcurrentQueue<string>();
         private readonly AutoResetEvent _messageEvent = new AutoResetEvent(false);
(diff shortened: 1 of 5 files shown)


src/Wyam/Program.cs
```diff
@@ -394,6 +396,10 @@ private Engine GetEngine()
                 {
                     engine.CleanOutputPathOnExecute = false;
                 }
+                if (_engineGlobalMetadata != null)
+                {
+                    engine.GlobalMetadata = _engineGlobalMetadata;
```
miere43: Engine's global metadata is not read-only, so in previous watch mode iterations this metadata may be changed. Maybe we should clone metadata for each engine instance or make metadata read-only.

daveaglick: That's a very good catch. Conceptually, we want the global metadata to start at the same point for each generation because any changes that get made will be "replayed" for following generations. If it persisted, our generations might end up polluting each other. I'd like to leave the global metadata writable from the config file too (it's already read-only from the execution context). That leaves some sort of cloning operation for each new engine.

`SimpleMetadata` is literally just a `Dictionary<string, object>`, so what if we added `ISimpleMetadata.Clone()` (which shouldn't be hard to write) and then made the line above:

```
engine.GlobalMetadata = _engineGlobalMetadata.Clone();
```

daveaglick:
:+1: This is a great bug fix, glad you found it. I made one line comment about cloning the metadata, let me know what you think.

miere43:
Seems to be okay now.

daveaglick:
Perfect!

Issue #127, opened 2015-11-19
Create a Windows installer


daveaglick:
Will make onboarding easier for some users who aren't used to the command line. Might also make a lightweight GUI just for picking folders and running (cross-platform? what GUI toolkit? Eto.Forms?) Should also install service described in #21

FlorianRappl:
Let's just use [Squirrel](https://github.com/Squirrel/Squirrel.Windows) for that. I would write the installer and make the PR.

We could have the application self-updating using a new boolean flag `--update`.

Should I start working on it?

daveaglick:
> Should I start working on it?

Yes! That would be awesome.

This is one of those issues that I think is really important for adoption, but I just haven't had the time to dig into yet, so taking it on would be greatly appreciated. And I love the concept behind Squirrel, so I think that's a really good choice.

PR #283, opened 2016-05-04
Integrated Squirrel as Installer, see #127

FlorianRappl:
This is part of the installer from #127. Currently, the following things have been integrated:
- Auto updating via the `--update-self` command flag (takes last release from GitHub)
- Ability to build an installer via the `Create-Installer` target
- Publish the installer on the `Publish-Release` target

The latter has not been tested yet - maybe you can do that with the next release. Also the updating is something that needs to be tested productively.

The following things are currently missing:
- The installation does not add the application to the `PATH` (see below)
- Previous releases are not gathered before a new build resulting in full "updates" instead of delta updates

In order to be really useful the `wyam.exe` should be available in the path, however, there is no option for Squirrel to provide this. The only way is to provide our own `Setup.exe` template, but there I see two problems:
1. I dislike providing a binary
2. What should the path be? The `wyam.exe` path switches from version to version; only the Squirrel installer knows what it removes / adds. It uses this information, e.g., when creating shortcuts.

My idea to tackle this problem would be to suggest a new feature for Squirrel, where the `ShortcutLocation` can be set to "Path", such that the application is added to the path. This would be beneficial for command line applications such as Wyam. The risk is of course that this feature is disliked / not wanted.

What are your thoughts?


daveaglick:
Whoa, awesome! I won't get a chance to look it over until tomorrow, but wanted to drop in and say thanks. This will really help with adoption and improve the experience once it goes live.

build.cake
```diff
@@ -233,6 +235,32 @@ Task("Create-Tools-Package")
...
     });
+
+Task("Create-Installer")
+    .IsDependentOn("Copy-Files")
+    .Does(() => {
+        CopyDirectory(binDir, installerDir + Directory("lib") + Directory("net45"));
+        CopyFileToDirectory("./src/Wyam.nuspec", installerDir);
+        var nuspec = GetFiles(installerDir + File("Wyam.nuspec")).FirstOrDefault();
+
+        NuGetPack(nuspec, new NuGetPackSettings
```
daveaglick, line 246: Do we need to create a new package for wyam.exe here? We're already creating one using the nuspec at src\Wyam\Wyam.nuspec up in the Create-Tools-Package task. Could the Create-Installer task depend on Create-Tools-Package and use that resulting package? In other words, curious what the difference is between the "tools" package and the one we create for Squirrel?

build.cake
```diff
@@ -179,22 +182,21 @@ Task("Create-AllModules-Package")
...
-        }
+        List<NuSpecDependency> dependencies = new List<NuSpecDependency>(
+            GetFiles("./src/**/Wyam.Modules.*/*.nuspec")
+                .Where(x => x.GetDirectory().GetDirectoryName() != "Wyam.Modules.All")
+                .Select(x => new NuSpecDependency
+                    {
+                        Id = x.GetDirectory().GetDirectoryName(),
+                        Version = semVersion
+                    })
+        );
```
daveaglick, line 199: :+1:

build.cake
FlorianRappl, line 246: The one that is packed here contains everything that would otherwise be distributed in the zip / bin directory. The one you are referring to (but maybe I got that wrong) is referencing the files from the `bin` folder, but as they are not placed in `lib/net45` Squirrel won't accept them. But potentially I've missed something. Having one package would be great, but I did not want to disrupt your current setup too much.

daveaglick, line 246: Ah, got it. That makes sense. For a "tools" package (I don't know if that's the official name for these or not), you put everything under /tools. The NuGet clients know that if there's stuff in /tools that it should do a little magic and make it available for use in the Package Manager Console.

I've never actually used Squirrel - so found this a bit interesting when going through their docs just now:

> Squirrel is expecting a single lib / net45 directory provided regardless of whether your app is a net45 application

I may suggest/submit a PR to allow that to be customized or to use /tools as a fallback. In the meantime let's just leave this as-is since it works.

src/Wyam/Settings.cs
```diff
@@ -49,6 +50,7 @@ public bool ParseArgs(string[] args, out bool hasErrors)
                 syntax.DefineOptionList("i|input", ref InputPaths, DirectoryPath.FromString, "The path(s) of input files, can be absolute or relative to the current folder.");
                 syntax.DefineOption("o|output", ref OutputPath, DirectoryPath.FromString, "The path to output files, can be absolute or relative to the current folder.");
                 syntax.DefineOption("c|config", ref ConfigFilePath, FilePath.FromString, "Configuration file (by default, config.wyam is used).");
+                syntax.DefineOption("self-update", ref SelfUpdate, "Tries to find and install an updated version of the application.");
```
daveaglick, line 53: I love that updating will be as easy as just adding a CLI flag. The System.CommandLine library I'm using supports the concept of [_commands_](https://github.com/dotnet/corefxlab/tree/master/src/System.CommandLine#defining-commands), or different operating modes for the application. Perhaps the update process should be it's own command since you probably wouldn't want to both update and run a build at the same time. That would also allow us to eventually pass update-specific flags like version (don't know if Squirrel supports this), source, etc.

Hopefully System.CommandLine supports a default command and the current set of options can still be provided as such - if I have to differentiate between the update command and the build command on every execution, I'd rather keep it as a flag like it is here.

daveaglick:
Thanks again for taking this on. It looks good and I tested it locally with success. I think my biggest question at this point is how it will affect porting to .NET Core (or even Mono or .NET Native when that's ready). I'd love to get Wyam cross platform at some point in the near future. Obviously this change doesn't touch the core or common libraries, but I'd also like to use the same CLI code if possible too.

Given the set of dependencies Squirrel requires, I'm guessing not all of them are going to be Core-ready. I'm wondering if it's possible to integrate Squirrel specifically for Windows while not precluding using the Wyam CLI on other platforms. Right now, I've got three ideas of how we might accomplish this:
- The use of preprocessor directives and conditional references.
- Create a separate Windows-specific updater/bootstrapper application that can be responsible for managing the update process.
- Split most of the CLI functionality into a separate general xplat library project that has mechanisms for adding additional CLI options/commands/etc. Then create platform-specific console applications that each reference the common CLI library.

Right now I'm favoring the last option because it seems very flexible. I.e., it would let us add platform-specific CLI commands however is appropriate for each platform. If that's the case we can accept the PR as-is and I'll do the split later when it becomes more obvious what the xplat strategy looks like.

Thoughts?

(Also, I'm still thinking about the path problem you mentioned above - I agree it should be solved, getting wyam.exe on the path is really important if we're going to be changing where it's located all the time).

FlorianRappl:
For a first release I wouldn't worry about the path problem. Adding the application to the path can be done in a single step and is no blocker (but it is certainly something to include going forward).

If you want to make the CLI x-plat then 3.) is definitely the way to go (long run). However, regarding the special case of Squirrel itself I would say 1.) solves the problem without much trouble.

3.) would have other benefits as well, such as having a web frontend or GUI - crazy things to think about, but potentially useful in certain environments (imagine a Wyam service in an intranet for instance). It's nothing to write / care about at the moment, but having the flexibility is certainly beneficial.

daveaglick:
I'm going to merge this in shortly. I gave this change _a lot_ of thought over the weekend. The part that made me uncomfortable was how Squirrel, a Windows-only library that isn't really related to the core responsibility of the application, would become a hard dependency. The more I thought about it, the more I also didn't like the idea of bifurcating the CLI into multiple platform-specific applications. I find value in the actual .exe being (eventually) cross-platform and I don't want to move away from that.

So after I merge this, I'm going to make a couple changes:
- Move all the Squirrel code into a separate `Wyam.Windows` application that will serve almost exclusivly to drive the installer (at least for now).
- The Squirell NuGet package will include both the `Wyam.Windows` and the cross-platform `Wyam` outputs side-by-side.
- This new `Wyam.Windows` application will also handle [Squirrel events](https://github.com/Squirrel/Squirrel.Windows/blob/master/docs/using/custom-squirrel-events.md).
- On each update it will (re)create `wyam.cmd` and `wyam.windows.cmd` in the top-level folder which will just proxy the command line args to the .exe in the latest version folder.
- To update (or do any other future Windows-specific management), the user will call `wyam.windows update`. I envision it being used in the future for other Windows-specific management like `wyam.windows install-service`.

Thanks again for this! It's getting really close to another release and having this along with the new Cake.Wyam addin for Cake makes a really compelling story around deployment.

FlorianRappl:
This is great! :+1: to the Squirrel events. `wyam.cmd` to proxy calls is the way to do it (at least this is the way Node.js does it and this works pretty well for all my tools, may they be external or internal).

Issue #240, opened 2016-02-21
Breaking change with 0.11.2-beta - 'IDocument' does not contain a definition for 'Clone'


avishnyakov:
Having the following after migration to 0.11.2-beta:

```
Line 80: CS1061: 'IDocument' does not contain a definition for 'Clone' and no extension method 'Clone' accepting a first argument of type 'IDocument' could be found (are you missing a using directive or an assembly reference?)
```

Can't see that mentioned in https://github.com/Wyamio/Wyam/releases/tag/v0.11.2-beta 

How'd we overcome it? The following line raises the issue:

``` cs
// custom exec andler
Execute( (doc, context) => {

var result = new List<IDocument>();

// bla bla bla bla
// process HTML doc

var newDoc = x.Clone(htmlDocument.DocumentElement.OuterHtml);
result.Add(newDoc);

return result;
}),
```

daveaglick:
Aw, crap. I could swear I added a release note about this - I'll stick one in tomorrow when I'm back at a desktop.

`IDocument.Clone()` was removed as part of switching to factory pattern for documents (which in turn enables some advanced scenarios like custom document types). You just need to replace those calls with `IExecutionContext.GetDocument()` (the first parameter is the document to clone).

```
var newDoc = context.GetDocument(x, htmlDocument.DocumentElement.OuterHtml);
```

daveaglick:
Sorry about that. The next couple releases will have some breaking changes as I refactor some internals to support recipies and themes with a more advanced IO abstraction. I'm trying to keep it to a minimum, but there may be some more like this one. I'll try to be extra careful to point them out in the release notes.

avishnyakov:
Works well!

PR #288, opened 2016-05-17
Changed Remarks of IDocument to point to the location of Clone

LokiMidgard:
Hi,

the Remarks read like following:
"Documents are immutable so you must call one of the **Clone** methods to create a new document. ..."

I added a link to IDocumentFactory
I think this is related to #240 

Not sure if my change is ok. Languages are not my best subject.


daveaglick:
:+1: I love little doc PRs like this because it's so often stuff that gets missed during refactoring

Issue #150, opened 2015-12-18
New module to create a tree from documents in pipeline


daveaglick:
This would be helpful in a lot of different use cases where the documents in the pipeline naturally make a tree and you want to process them that way. For example, the current ReadFiles module reads all the documents for a file system tree into the output document sequence. We then rely on a metadata value to tell us where the document sits in the tree. Other modules like CodeAnalysis also have a tree-like relationship but use metadata to represent the tree (I.e., namespace relationships).

On the other hand, we've seen in the CodeAnalysis module how storing sequences of documents within a container document can work very well. I'd like to adopt this strategy for this new module. Here's how it should work:
- You specify a metadata value to look at along with a delimiter. For example, to make a tree from documents output by ReadFiles, you would use the SourceFilePath metadata value and a path separator delimiter.
- The module will split the specified metadata value (as converted to a string) for each input document according to the specified delimiter.
- A new Children metadata value (or other user-specified metadata key) will be added to all documents in the tree that have child documents. This contains an `IReadOnlyList<IDocument>` of all child documents in the order in which they appear in the input document sequence.
- New documents with empty content will be created for roots and branches in the tree that don't have an existing document to represent them. I.e., when processing output from ReadFiles, all the input documents will be files so new empty documents for each directory will end up being created.

A new `ReadFiles.OutputTree()` method should also be added to automatically output the result from ReadFiles in this way rather than the default flat output sequence.

avishnyakov:
Fair enough, all makes sense.

User stories would not be only about metadata but also around folder structure.
In Wyam web site the ToC is built around metadata and groups (comes from metadata too), however building ToC based on the folder structure is another desired case.

Having said that, there are also hybrid cases: a tree is constructed in highly custom ways partly with metadata, partly based on the folders structure, index.html files, partly with the fallbacks (if no metadata, then use filename) and so on.

All that leads to the idea of having not only default ReadFiles.OutputTree() extension, but also a callback so that you can own the lookup logic. Something like "ReadFiles.OutputTree(config)" where the config would be a set of functions to customize the lookup logic.

Here is a SO discussion about "Nice & universal way to convert List of items to Tree":
http://stackoverflow.com/questions/19648166/nice-universal-way-to-convert-list-of-items-to-tree
"GenerateTree" method seems just about right, does not it?

LokiMidgard:
[...]

Issue #277, opened 2016-04-22
Refactor file providers to use a URI


daveaglick:
A problem I keep coming back to is how to "define" a given file provider. First thought was that it would be done in code and a given `IFileProvider` instance would just be constructed and added to a collection of available providers. This has some downsides:
- It requires the user to be familiar with the various constructors for each file provider.
- It requires the code for constructing and adding the provider to be evaluated before the provider is used.
- It makes it hard/impossible to instantiate file providers on the command line for scenarios without a configuration.

A good solution to all these problems would be to use URIs for specifying the file provider. The path portion should remain the same - the URI will just be used for the provider. Extra parsing will be needed to handle string to path conversions where the string is a URI (the scheme, hostname, etc. portion should be the provider and the path and query should be the path).

Since `:` has meaning in URIs and many other ASCII characters are explicitly reserved or have accepted meaning in URIs, we need a new delimiter between the file provider and the path when specifying paths as strings. The `|` character seems like the safest bet - it's unsafe in URIs and generally needs to be escaped (see [RFC 1738](http://www.ietf.org/rfc/rfc1738.txt) 2.2, though [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) no longer includes them) and it doesn't already have some other meaning for common URI schemes. Thankfully, since Wyam hasn't implemented other file providers there's been no need to use the previous delimiter and this has a very low chance of being a breaking change.

The following path representations should all be equivalent:
- `"gist|/2721371adf6bf69fa833"` - provider portion is implicitly understood to be a scheme without a host or any other components if not a URI (translates to "gist:///").
- `"gist:///2721371adf6bf69fa833"` - if no provider is specified explicitly but the path itself is a valid URI, then the path and query become the path and the scheme, host, etc. become the provider. Note that in this example, `///` is used and not `//` because a double slash would have indicated the gist ID was a host name and it would have ended up in the provider with an empty path.
- `"gist://|/2721371adf6bf69fa833"`
- `new FilePath("gist", "/2721371adf6bf69fa833")`
- `new FilePath("gist://", "/2721371adf6bf69fa833")`

Here is an example with host names where the paths are equivalent:
- `"http://foo.com|/a/b/c.txt"`
- `"http://foo.com/a/b/c.txt"`
- `new FilePath("http://foo.com", "/a/b/c.txt")`

And here is another more complex example that doesn't use host names but does require a path as part of the provider:
[...]

PR #292, opened 2016-05-20
Added Tree Module

LokiMidgard:
Created a Tree Module (as in #150), that adds Metadata to each document that represents the tree structure.
The Module takes a DocumentConfig that returns for each Document a `string[]`. This is used to generate the tree.

Every Document get Metadata added to navigate around.

In addition for each Document a bool can hint to split the Tree on that node. Removing the node from the original tree and creating an additional tree with the node as root.

Actually I'm not sure if the last thing is needed. I think it could also be archived when manipulating the path that describs the tree structure.

### DIFF
--- a/src/Wyam.Core.Tests/Modules/Metadata/TreeMetaTests.cs
+++ b/src/Wyam.Core.Tests/Modules/Metadata/TreeMetaTests.cs
@@ -1,0 +1,762 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using NSubstitute;
+using NUnit.Framework;
+using Wyam.Common.Documents;
+using Wyam.Common.IO;
+using Wyam.Common.Meta;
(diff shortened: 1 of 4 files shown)


daveaglick:
This looks really cool, and I especially like how you can supply a path function so that you can construct an arbitrary tree structure if needed.

Any chance you could add a few tests? That would help not only with verifying functionality, but perhaps more importantly, serve as rough documentation.

LokiMidgard:
Well took some time to create those tests.

daveaglick:
Those are awesome! Thanks for putting them together. Makes it so much easier to learn how it all works. I tend to point folks towards the tests as an initial "here's how you use it," so this should help with onboarding.

I'll hopefully get this pulled in tomorrow. I'm hoping to get another release out next week now that the changes in #277 are nearly done too (don't worry, no breaks this time around :smile:).

PR #294, opened 2016-05-20
Fix nullpointer Exception if stream of Document is null

LokiMidgard:
When creating a Document using `ExecutionContext.GetDocument()` the `_stream` field of the document is null. Calling the `ToString()`-Method then results in a `NullReferenceException`
Now it returns an empty string instead.


daveaglick:
:+1:

PR #306, opened 2016-05-31
Sidecar file

LokiMidgard:
Implementing a Module for [Sidecar Files](https://en.wikipedia.org/wiki/Sidecar_file).

This allows to add Metadata to Documents without modifying the content of the document. The Metadata for the File `Test.xml` would be stored in `Test.xml.meta`.

Like FrontMatter this module dos not process Metadata itself. It uses other Modules to do this.

This Module can be usefull with binary formats or text files that can not add FrontMatter without lousing editor support for autocompletion or syntaxhilighting.


daveaglick:
:+1: This is a great idea - I know some other static generators support sidecar files, so I'm thrilled to get it into Wyam. Thanks!

PR #321, opened 2016-06-21
Fix dependency on broken version of Nuget.Packaging.Core

M-Zuber:
See https://github.com/NuGet/Home/issues/2690 for reference

### DIFF
--- a/src/Wyam.Configuration.Tests/app.config
+++ b/src/Wyam.Configuration.Tests/app.config
@@ -4,15 +4,23 @@
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
         <assemblyIdentity name="NuGet.Versioning" publicKeyToken="31bf3856ad364e35" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-3.4.3.0" newVersion="3.4.3.0" />
+        <bindingRedirect oldVersion="0.0.0.0-3.5.0.0" newVersion="3.5.0.0" />
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="NuGet.Frameworks" publicKeyToken="31bf3856ad364e35" culture="neutral" />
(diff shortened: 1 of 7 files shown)


daveaglick:
Quick question - when you got the strong name error this morning, were you building/running from source or using the pre-packages Wyam zip file or Windows installer?

M-Zuber:
Source

daveaglick:
Okay, good - that probably means that the issue with the 3.4.3 package was actually isolated to your own system. They pretty quickly replaced it with a working one, but you must have restored packages at just the right moment when the broken one was up. I misunderstood and thought you were working from the official release of Wyam.

My hesitation with this pull request is that it uses pre-release NuGet packages - after checking, it looks like 3.4.3 is still the most recent stable release. Their non-pre-release stuff is already flaky with all the changes going on the the ecosystem right now. I'd be worried that a pre-release NuGet package would just add to that. Since the bad 3.4.3 package has already been replaced on the gallery feed, I'm not sure the version bump is actually needed.

I'll leave this open for a while in case it turns out there's a bigger problem, but in the meantime I'll hold off to see if they release a production package soon.

M-Zuber:
That would be odd, as I opened the source on my machine for the first time 2 days ago - which should have been after the fix.

I hope to get my new machine set up by the end of the week, and then I can try again on a clean machine. Will keep you posted

daveaglick:
Let me know if everything works on a clean system... I tested a fresh install on a VM and it appeared to compile and run okay, so I'll close this once I get the all-clear from you.

daveaglick:
I'm going to go ahead and close this. I've tested on multiple clean environments and haven't seen any additional problems since the initial report.

M-Zuber:
Fair enough, apologies I never got back to you on this. 
My clean environment that I thought I was going to have setup is still a work in progress due to all sorts of issues :(

daveaglick:
No problem :smile:

Good luck getting everything setup, I know you've been working on it for a while...

M-Zuber:
Someday when it's not 12:30am, I'll tell you the story 😁

PR #328, opened 2016-08-01
Markdig

### LINKED ISSUE #165: Consider changing Markdown processor to Markdig
The current one is CommonMark implementation, so there aren't any neat familiar github-flavored markdown goodies. It would be nice to see alternative markdown implementations with extended syntax support.

FlorianRappl:
Regarding #165 I did the following:
- Replaced CommonMark.NET with Markdig (using the default HTML Renderer)
- Extended the Markdown extension (pun intended) with Markdig specific methods (to fully utilize it for further use cases)
- Adjusted the unit tests (all of them had 1 more new line - which was specific to CommonMark.NET; however, now we need to use `\n` instead of the system's newline - this is specific to Markdig)*

*) This is pretty much hardwired; we could change it by self-creating the `HtmlRenderer` instance with a `TextWriter` that does use the environment's newline sequence. However, I thought being cross platform also means that `\n` should be preferred instead of the Windows specific `\r\n` being produced on Windows.

There could be more points of extension, e.g., allowing the user to use other renderers or making the pipeline more open (right now one may choose between all of the available extensions and include new extensions, but the the core parsers cannot be modified).

Nevertheless, the PR should be sufficient to close #165 - more additions should be handled in more issues (and specified with all requirements).

### DIFF
--- a/src/Wyam.Markdown.Tests/MarkdownTests.cs
+++ b/src/Wyam.Markdown.Tests/MarkdownTests.cs
@@ -41,6 +41,96 @@ namespace Wyam.Markdown.Tests
             }
 
             [Test]
+            public void DoesNotRenderSpecialAttributesByDefault()
+            {
+                // Given
+                string input = @"[link](url){#id .class}";
+                string output = @"<p><a href=""url"">link</a>{#id .class}</p>
+".Replace(Environment.NewLine, "\n");
+                IDocument document = Substitute.For<IDocument>();
+                document.Content.Returns(input);
+                IExecutionContext context = Substitute.For<IExecutionContext>();
+                Markdown markdown = new Markdown();
+
+                // When
+                markdown.Execute(new[] { document }, context).ToList();  // Make sure to materialize the result list
+
+                // Then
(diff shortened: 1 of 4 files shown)


daveaglick:
This is fantastic! I've got no problem at all with the newline difference - as you mentioned, `\n` is probably  more standards conformant anyway. I'll take a look and merge this probably tomorrow. I'd like to get a new version released within the next day or two and would _love_ to get this released with it.

FlorianRappl:
I guess I'll add one or two more unit tests and one more example. I don't know if I find the time today though. If you want to merge - go ahead (the unit tests will cover the extensions API, but that seems to be alright anyway; nevertheless, I want to cover this functionality for the future).

daveaglick:
Thanks a ton - works like a charm! Already getting ideas for cool Markdown extensions...

FlorianRappl:
I have one that can replace variables in there; right now I see two options for this. Either the variables are static or the variables are being evaluated by some kind of scripting engine (e.g., Roslyn, or my own pet project MAGES). Since Roslyn is loaded right away it would make sense, but I have to dig deeper into this.

In any case this is potentially an addon that is decoupled from the main thing. I need to look at specific requirements first.

PR #340, opened 2016-09-28
Added Xslt module that uses saxon

LokiMidgard:
I created an alternative Xslt (Xslt2) module that does use saxon instead of microsofts xslt processor. This allows the use of XSLT 2.0. I keept the original Xslt module because it is not depandend on 3ed party libs.


daveaglick:
Cool, thanks! Looks like there's some CI errors, so I'm going to try and clear those and then I'll merge it in.

daveaglick:
@LokiMidgard FYI, I'm planning drop the Xslt2 module. It's not `netstandard` compatible because the Saxon libraries all target .NET Framework (and I don't see that changing soon since they're using IKVM). Not to mention, [IKVM itself will no longer be updated](http://weblog.ikvm.net/) so there's no chance of it becoming `netstandard` compatible. And finally, the reference in the Saxon NuGet package isn't compatible with SDK-style projects so getting the Saxon library into the module project is now a problem.

All this adds up to a huge support burden for a niche module given that we also have an Xslt module. If/when [corefx adds support for XSLT 2.0](https://github.com/dotnet/corefx/issues/2295) or an alternate `netstandard` library is available we can bring it back without Saxon.

LokiMidgard:
I don't have a problem with that. But I hope that corefx will add this in the near future.

If I find time, I will try to create a seperate project/nuget pacakge that someone who needs XSLT 2.0 and and runs on .Net Framework can use. At least untill it is supported by netstandard.

PR #355, opened 2016-11-05
Fixed bug that BaseUri was not always set in Xslt2 module. 

LokiMidgard:
Only when _xsltGeneration was used the BaseUri got set. using _xsltPath did not set this. This could result in an exception even when the BaseUri is not needed.


daveaglick:
Somehow totally missed this getting submitted while I was away the last couple weeks - sorry!

LokiMidgard:
No problem. if it wouldent merged untill the next release I would have asked ;)

Issue #371, opened 2017-01-04
Wyam.windows has wrong version of System.Console


adamclifford:
When attempting to add wyam to the %PATH% using the wyam.windows.cmd the following error is displayed :-
```
wyam.windows.cmd update
Wyam version 0.15.6-beta
System.IO.FileLoadException: Could not load file or assembly 'System.Console, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
File name: 'System.Console, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
   at System.CommandLine.ArgumentSyntax.Validate()
   at System.CommandLine.ArgumentSyntax.Parse(IEnumerable`1 arguments, Action`1 defineAction)
   at Wyam.Windows.Settings.ParseArgs(String[] args, Boolean& hasErrors) in E:\Code\Wyam\src\clients\Wyam.Windows\Settings.cs:line 0
   at Wyam.Windows.Program.Run(String[] args) in E:\Code\Wyam\src\clients\Wyam.Windows\Program.cs:line 43
   at Wyam.Windows.Program.Main(String[] args) in E:\Code\Wyam\src\clients\Wyam.Windows\Program.cs:line 24

WRN: Assembly binding logging is turned OFF.
To enable assembly bind failure logging, set the registry value [HKLM\Software\Microsoft\Fusion!EnableLog] (DWORD) to 1.
Note: There is some performance penalty associated with assembly bind failure logging.
To turn this feature off, remove the registry value [HKLM\Software\Microsoft\Fusion!EnableLog].
```

To fix I copied the assembly binding redirect from waym.exe.config

```
<dependentAssembly>
        <assemblyIdentity name="System.Console" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.0.1.0" newVersion="4.0.1.0" />
</dependentAssembly>
```

daveaglick:
Resolved by #372

PR #372, opened 2017-01-04
Fixing issue #371

adamclifford:
adding in the correct assembly bindingd


daveaglick:
Thanks! These binding redirects for the new netstandard libraries to retarget away from the older framework ones in the GAC are killing me lately. Thanks for catching this one.

PR #384, opened 2017-01-13
Added CsvToMarkdown module

LokiMidgard:
I created a modification of the CsVToHtml Module rendering to Markdown tabels instead html.

### DIFF
--- a/src/extensions/Wyam.Tables/CsvToMarkdown.cs
+++ b/src/extensions/Wyam.Tables/CsvToMarkdown.cs
@@ -1,0 +1,101 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Wyam.Common.Documents;
+using Wyam.Common.Modules;
+using Wyam.Common.Execution;
+using Wyam.Common.Tracing;
+
+namespace Wyam.Tables
+{
+    /// <summary>
+    /// Converts CSV content to Markdown tables.
+    /// </summary>
+    /// <remarks>
+    /// This module reads the content of each input document as CSV and outputs an Markdown table 
+    /// containing the CSV content. The input CSV content must use <c>,</c> as separator and enclose 
+    /// every value in <c>"</c>. The output table has the format
+    /// 
+    /// +--------------+-------------+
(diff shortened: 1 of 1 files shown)


daveaglick:
Cool idea, and I'm not opposed to pulling it in. I'm just wondering about the use case though. HTML is kind of a universal output since it's directly valid in almost every templating language including Razor, Liquid, and Markdown. Since HTML is itself valid Markdown, what does outputting tables in Markdown syntax buy that outputting as HTML and including that in a Markdown file doesn't?

LokiMidgard:
**tldr:** 
 + A use case is when markdown is not only an intermediate format. Then using html is not a good option. ```<table>...</table>``` is not as fancy as a table created from ```+``` and ```-```.
 + I have problems separating design and structure in html.

----------------------------------------------------------------------
**Long version**

Well it is true that markdown includes html, but it does not "feel" like markdown. There are two parts.
 + I currently thinking about not only rendering the content of my site in html as a website but other formats. The two I have in mind are pdf and a rfc like text version, which would more or less be the created markdown files (but only if no html tags are included).

And the for me personal more important part
 + psycolgical effect when I creating the source documents and the pipelines that orchestrate everything. When I start including html in markdown I automaticly start thinking website. And when I start thinking Website I start thinking with things like bootstrap and such in mind. Now I start to create extra divs around divs and those need to special classes and id's that will later be used in the JavaScript. It is no longer a clear separation of the content and the design I want later to apply in the Website.

Restricting me to no longer using html in the content part of my site helped me to better structure the build pipelines No longer mixing design in the structure of the content. And I hope that will help me later on when I actually start to not only create the website but the PDF/Text/whatever with the same source Pipeline.

daveaglick:
Okay, that makes sense - especially the use case where Markdown may be consumed directly without rendering to HTML. I'll merge this in prior to the next release.

I do think we need to tread a little carefully here. I want to be careful not to get in a mindset of creating modules to handle every possible input/output combination. CsvToLiquid, ListToMarkdown, etc. This particular case is niche enough that I'm not worried, but just something I'll have to keep in mind as we continue to expand the set of official modules. But don't get me wrong - more useful modules are great! Just want to be sure it doesn't get too confusing.

PR #397, opened 2017-01-20
Support RssPath and AtomPath in themes (#396)

### LINKED ISSUE #396: CleanBlog not using RssPath or AtomPath metadata
I was trying to change the Atom XML path and noticed the theme has hardcoded the links to the RSS/Atom feeds:

https://github.com/Wyamio/Wyam/blob/develop/themes/Blog/CleanBlog/_Footer.cshtml#L7

Also in Layout:

https://github.com/Wyamio/Wyam/blob/develop/themes/Blog/CleanBlog/_Layout.cshtml#L16

kamranayub:
Closes #396 

## Changes

- Add support for `RssPath` and `AtomPath` in `_Layout.cshtml` and `_Footer.cshtml` for CleanBlog and Phantom themes


daveaglick:
Big :+1: for this, thanks again!

kamranayub:
One thing I didn't test, whether this will work for a path without a
leading slash (relative) since the previous strings had an absolute path.
Do you know?

daveaglick:
It should be okay - the `.GetLink()` call is supposed to account for those differences. Thanks for the heads up though, I'll make sure to double-check this case.

kamranayub:
Just checked, it seems to work fine :)

PR #398, opened 2017-01-20
Allow a filter function for CopyFiles.

deanebarker:
This is the idea I have -- allow a "last minute" filtering function to alter the destination path for files. This would happen _after_ Wyam has gone to all the trouble of calculating path relativity.

I struggled a bit on naming. It might suck.

(Also, I'm not a developer...)


daveaglick:
Ah, okay, now I get where you were going with this. I do have a suggestion - I think it would be clearer what's going on (and simplify the module call) if we just made this an overload:
```
To(Func<IFile, IFile, FilePath> destinationPath)
{
}
```

That way, you can have access to both the source `IFile` and the calculated destination `IFile` at the same time if needed (or just ignore the source `IFile` if you don't care). I'll make the adjustment post-merge, but want your feedback - we can always go back and change it again or change it back to something closer to the original PR.

PR #400, opened 2017-01-21
Use PostsPath for Archive links in themes

### LINKED ISSUE #399: PostsPath not used for Archive link on Index
Similar to #396 

The `PostsPath` property is not used on `_Index.cshtml` for the Archive link:

https://github.com/Wyamio/Wyam/blob/develop/themes/Blog/CleanBlog/index.cshtml#L57

Not sure how Archive link is generated in navbar--that may be working.

**edit:** Actually, maybe this isn't an issue--I think it expects the archive path to always be at `/posts` no matter where the posts are sourced from.

kamranayub:
Closes #399 

## Changes

- Use `Context.String(BlogKeys.PostsPath)` for Archive link generation in themes


kamranayub:
Not sure if this will be an issue too?

https://github.com/Wyamio/Wyam/blob/develop/src/extensions/Wyam.Blog/Blog.cs#L49

Looks like that may be okay since it's the default theme path, then it gets overridden by Meta module...

Actually, now that I look at this--maybe there isn't an issue, if the Archive index is always in the same location?

daveaglick:
Thanks!

PR #412, opened 2017-01-24
Remove top margin from paragraph when in blockquote

n-develop:
When a blockquote is used inside a blog post using CleanBlog theme, it shows a top margin. This margin is useful for single p-tags. But inside the blockquote-tag it is not needed.

Currently, it looks like this:
image

This simple CSS change makes it look like this:
image


daveaglick:
And a second PR!

n-develop:
Cool. :sunglasses:  Thanks again.

PR #418, opened 2017-01-27
Added missing Module to project

LokiMidgard:
It seems I missed to commit the Project file when I created the Module. It is not included in the current release ( which is no longer beta (^- ^ )).

Sorry for this.


daveaglick:
Whops! We'll get it out on the next release.

Issue #421, opened 2017-01-28
When Watching, Execution Errors should not Kill Wyam


Silvenga:
I'm currently theming my Blog created using Wyam using Razor and whenever I make a syntax mistake I have to restart Wyam. It would be nice to mimic how Grunt or Webpack handle this - when compilation/rendering fails while watching, wait until more changes are detected and attempt to compile/render again.

daveaglick:
Totally agree.

PR #422, opened 2017-01-28
Prevent ExecutionError's from killing the activity loop

Silvenga:
See #421.


daveaglick:
Thanks!

The only concern I have here is that the persistent cache that sticks around from one generation to another doesn't get messed up. I'm guessing it's fine, but want to remember to take a look at that when I review.

daveaglick:
Thanks a lot for this, should make iterating on a new site much faster. Also, love the branch name `mercy-on-mistakes` :smile:

Silvenga:
Sweet!

Issue #323, opened 2016-07-01
Brotli or other image compression module


schulz3000:
There are still good modules for CSS, Javascript and HTML minimizing.
It will be good to have an image shrinking module to save bandwidth.

Also it should be considered if shrinking of images must be always lossless or if there is an option needed for lossy shrinking

miere43:
You can use [Image](http://wyam.io/modules/image) module, there are `Resize` method lying in there. I don't know what is _lossy shrinking_ because when you make image smaller, you losing pixels anyway.

daveaglick:
At some point, it would be cool to get a Brotli compression module (or to work it into the Image module). The tradeoff is build speed though, I've heard Brotli can take a while.

n-develop:
I use the jpg compression of the existing Image module. Is does quite a good job. It shrinks my almost 1MB header image to less than 160KB without loosing a lot of quality.

Therefore, I use the following pipeline:
```csharp
Pipelines.Add("Images",
    ReadFiles("**/*").Where(x => new[] { ".jpg", ".jpeg" }.Contains(x.Path.Extension)),
    Image().SetJpegQuality(70),
    WriteFiles()
);
```
Even if 70 is the default quality, I find this a little more explicit.

Sadly, it just compresses jpg files.

petrsvihlik:
@daveaglick let's revive this one :)

I was playing around with Blazor recently and I noticed that after migration to .NET 5, [Brotli and Gzip](https://docs.microsoft.com/en-us/aspnet/core/performance/response-compression?view=aspnetcore-5.0) files started being generated out of the box. It occurred to me that it would be great to support these output types in Statiq too. It shouldn't be that difficult to add a possibility of configuring the type of output stream, right? There is [`BrotliStream`](https://github.com/dotnet/aspnetcore/blob/e3fd7471db53a559e37ba03dd76408053a80535e/src/Middleware/ResponseCompression/src/BrotliCompressionProvider.cs) included in ASP.NET Core out of the box. All of the major SSGs (Gatsby, Nuxt, Next, Gridsome) already have plugins for Brotli.

Or is it already somehow possible to inject an own implementation of `IFile` and output compressed files?

PR #427, opened 2017-02-01
Image guard clauses

n-develop:
The guard-clauses in the image module used "and" instead of "or".


daveaglick:
Great catch, thanks! It's unfortunate we've gone all this time with non-functioning checks (the Image module was one of the very first community contributions).

n-develop:
These things happen. It reads well in the first run. :wink:
I found it while I was trying to find out how this module works. Because I saw it uses the ImageProcessor library and thought this could be used for #323. Sadly compression just works well for jpegs.

PR #429, opened 2017-02-02
Adds additional extension to Markdown renderer to allow multiple markdig extensions

phil-scott-78:
Ran into an issue today with the extensibility of markdig when using recipes. I wanted to use Semantic-UI in a theme rather than bootstrap. All good until I started messing with tables. Semantic-UI wants a `ui table` css class on the tables similar to how bootstrap wants `table`. So I set about creating an extension for Markdig to support this. I could wire it up just fine, but to do so I had to override the recipe steps. The current setting only allows a string to be specified, and with this being custom it isn't in there obviously. So to get this custom extension added it involved copy and paste of both the `Pages` and `RawPosts` pipelines into the config file.

My proof of concept here adds a new setting for Blog and Docs named `MarkdownExternalExtensions`. It's an array of types. If nothing is set it continues as usual. Otherwise it applies them then applies the settings specified by the `MarkdownExtensions` property. The usage ends up looking like this

```
#n Markdig.SemanticUi
#recipe Blog

Settings[BlogKeys.MarkdownExtensions] = "advanced"; // get rid of default bootstrap
Settings[BlogKeys.MarkdownExternalExtensions] = new Type[] { typeof(Markdig.SemanticUi.SemanticUiExtension) };
```

I'm not a huge fan of the type array here, but short of letting Markdig leak all over the place I couldn't think of a cleaner way to do this so. Definitely up for suggestions.

Thoughts?

### DIFF
--- a/src/extensions/Wyam.Blog/Blog.cs
+++ b/src/extensions/Wyam.Blog/Blog.cs
@@ -38,7 +38,7 @@ namespace Wyam.Blog
             engine.Pipelines.Add(BlogPipelines.Pages,
                 new ReadFiles(ctx => $"{{!{ctx.DirectoryPath(BlogKeys.PostsPath).FullPath},**}}/*.md"),
                 new FrontMatter(new Yaml.Yaml()),
-                new Execute(ctx => new Markdown.Markdown().UseConfiguration(ctx.String(BlogKeys.MarkdownExtensions))),
+                new Execute(ctx => new Markdown.Markdown().UseExtensions(ctx.Settings[BlogKeys.MarkdownExternalExtensions] as IEnumerable<Type>).UseConfiguration(ctx.String(BlogKeys.MarkdownExtensions))),
                 new Concat(
                     // Add any additional Razor pages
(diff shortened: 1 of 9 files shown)


daveaglick:
I think this is a fine approach.

I can see why the type array isn't optimal - would have been cleaner to use generics or something, possibly with calls to `Markdown.UseExtension<TExtension>()` but the underlying challenge is how to pass data from the config file all the way down to modules in the recipe pipelines. The only alternative I can think of would be to add a static property or method to the `Markdown` module and then call that from the config file, but that's way worse than the type array IMO.

BTW, I appreciate the quality of this PR. I tend to be pretty lax about contributions and favor quantity over quality to help get folks involved (I just fix them up after if needed). It's clear you *get it* from the code here so I look forward to more in the future :wink:

phil-scott-78:
Cool, thanks for the kind words. I was mostly worried because any PR I see with `Activator.CreateInstance` I'll keep coloring ever so slightly outside of the lines here and see what else I can help out with.

PR #433, opened 2017-02-04
Fixes condition to look for nulls in phantom theme footer

phil-scott-78:
Using the phantom theme was throwing an exception rendering the footer
due to the lack of a proper conditional. Adjusts it to check for null.


daveaglick:
Thanks! Blow ups like this are embarrassing, glad you've got my back :smile:

PR #435, opened 2017-02-06
Uses BlogPipelines constants rather than BlogKeys

phil-scott-78:
This is a subtle one, but it threw me off when learning how pipelines and
the such work. `BlogKeys.Posts` and `BlogKeys.Tags` both are string constants
with the same value was `BlogPipelines.Posts` and `BlogPipelines.Tags`
respectfully. I was kind of turned around wondering the significance of
`BlogKeys` in relationship to the output of pipelines until I put it
together that the value is all that matters.

This simply swaps the constant being used and keeps the value the same so
there shouldn't be any breaking changes for people using the theme.

One thing of note - the phantom theme still has an error with a
conditional check in `_footer.cshtml`. PR #433 resolves this

### DIFF
--- a/themes/Blog/CleanBlog/_PostHeader.cshtml
+++ b/themes/Blog/CleanBlog/_PostHeader.cshtml
@@ -21,7 +21,7 @@
         <div class="tags">
             @foreach (string tag in Tags.OrderBy(x => x))
             {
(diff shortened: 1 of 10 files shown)


daveaglick:
Another good catch - there's been so much refactoring lately stuff like this tends to slip through the cracks.

Issue #348, opened 2016-10-27
Consider generation-time code highlighting (and diagrams)


daveaglick:
There's really no reason this needs to be done on the client other than it's easy to set up. A module that does code highlighting in a HTML file would be a nice alternative. Note that highlight.js can work via Node (https://highlightjs.org/download/) and that I recall seeing a .NET JavaScript/Node interpreter somewhere.

May also want to consider a similar approach for Mermaid diagrams.

daveaglick:
This was done a long time ago in the `HighlightCode` module using the now built-in JavaScript engine.

PR #442, opened 2017-02-10
Adds server-side code block formatting

phil-scott-78:
This applies the highlight.js highlighting on the server allowing the client side script to be dropped as mentioned in #348. It does this using the [MsieJavaScriptEngine](https://github.com/Taritsyn/MsieJavaScriptEngine) to execute the script. The script itself is actually the node version, but I applied browserify to it so that it wouldn't require node anymore. Kind of a hack, but it works. I included the steps I took to build the file in regenerating-highlight-all.js so that it could be updated in the future with a newer version.

Right now the main thing it is missing is the ability to add languages and some of the other highlight.js configuration options, but I'm not sure how many people are using it. Wanted to get this out there and get some feedback before going totally overboard with options. I noticed mermaid has an NPM package too - https://www.npmjs.com/package/mermaid. Potentially it could be applied in a similar fashion. I might give that a go too, but I'm not terribly familiar with that library.

One thing this did bring to light is how to slam a new module into an existing pipeline. I.E. I wanted to run this before the Razor module in the Posts pipeline so right now I have this ugly code

``` csharp
Pipelines[BlogPipelines.Posts].Insert(1, new Highlight());
```

Might be cool if the Pipeline allowed something like 

``` csharp
Pipelines[BlogPipelines.Posts].InsertBeforeFirst<Razor>(new Highlight());
```
or 
``` csharp
Pipelines[BlogPipelines.Posts].InsertAfterLast<Razor>(new Highlight());
```

### DIFF
--- a/Wyam.sln
+++ b/Wyam.sln
@@ -112,6 +112,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "integration", "integration"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Wyam.Docs", "src\extensions\Wyam.Docs\Wyam.Docs.csproj", "{0EB71E59-8D5C-45F3-BF1D-8649BA76D139}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Wyam.Highlight", "src\extensions\Wyam.Highlight\Wyam.Highlight.csproj", "{5FE758A2-1221-48FE-8A3C-8B3B9A32AFAB}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Wyam.Highlight.Tests", "tests\extensions\Wyam.Highlight.Tests\Wyam.Highlight.Tests.csproj", "{A6DE4586-1398-4071-90AB-5CCC1C949720}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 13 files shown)


phil-scott-78:
Mermaid might be a bit trickier. Seems they have some outstanding issues tied to it

https://github.com/knsv/mermaid/issues/260
https://github.com/knsv/mermaid/issues/169

daveaglick:
This is awesome - I didn't get a chance to look at it today, but will very soon.

Also, I *love*
```
Pipelines[BlogPipelines.Posts].InsertBeforeFirst<Razor>(new Highlight());
```
I've actually just been struggling with this myself on a large site and this would make it much easier (and more stable since an index would change if the underlying pipeline does). I'll open a separate issue for that addition (or you can if you beat me to it).

phil-scott-78:
@daveaglick - give me a bit on this one. I'm not a huge fan of having to build up a new `MsieJsEngine` for each document so I'm going to tweak it a bit to be more efficient. Running benchmarks right now for a baseline.

phil-scott-78:
I had an interesting issue the MsieJsEngine not liking being manually disposed, but only in AppVeyor. Going to investigate their library, but for now I've just removed the explicit `Dispose` for now. 

As to using `ThreadLocal<T>`, for a 1000 document site this went from 11s to 1s on my machine.

daveaglick:
I have the same concerns with this PR as I have with #439. I'm not sure MsieJavaScriptEngine gives us a good path forward for netstandard and cross platform compatibility. I wonder if the same approach used here with MsieJavaScriptEngine could be applied with [Jint](https://github.com/sebastienros/jint) instead, and how big an effort that would be?

phil-scott-78:
closing because #452 did this better.

Issue #443, opened 2017-02-11
Add enforcement for style rules


shaneray:
I was reviewing the `CONTRIBUTING.md` and came across the `Style Guidelines` section,  I was considering adding the `.editiorconfig` file from https://github.com/dotnet/roslyn/blob/master/.editorconfig to help stay consistent with accepted .NET coding styles.  Since this is something that is solution wide I wanted to open an issue/discussion to determine if this is something that would be beneficial.  Also up for discussion is any deviations from the roslyn `.editorconfig` that would be required/preferred for Wyam.

Silvenga:
In addition to `.editiorconfig` I propose an addition of a `.DotSettings` configuration as Resharper is used by a large portion of the C# developers market (and many cannot live without it).

shaneray:
I agree.  Another thing suggested by @enkafan is a simple set of style cop analyzers.  This would be useful for developers that are not using the latest visual studio with the `.editorconfig` support or ReSharper.  The more style helpers we can add to the projects the less chance style issues will make it into pull requests.  Also allows the style guidelines to be enforced by CI.

daveaglick:
Totally on board with this. The more tooling we can add to make hacking on the code easier and consistent, the better.

phil-scott-78:
Something else having the coding style standardized could lend itself to is integration with codebetter.io. I think they require a stylecop config though. Once that's in place it is quite nice to let it be the bad guy if someone submits a PR that doesn't color within the lines.

Tangent idea that might be worth a new issue  - could be clever if a module existed that could take an editorconfig and build out some text to include in the doc recipe.

phil-scott-78:
Err, codefactor not codebetter

daveaglick:
Reopening to track @enkafan idea of using CodeFactor or even just enforcing style rules by emitting build warnings (and treating as errors in CI).

shaneray:
@daveaglick I am currently working on some of these items.  I have added the .editorconfig and StyleCop.MSBuild nuget package to all projects.  Working on a few more tweaks before I submit my PR.

I also added my Wyam fork to CodeFactor.io and it is currently reciving a B grade.  I have added the badge below.  If you setup the original Wyam repository in CodeFactor I can add the badge to the readme in my PR.
[![CodeFactor](https://www.codefactor.io/repository/github/shaneray/wyam/badge/develop)](https://www.codefactor.io/repository/github/shaneray/wyam/overview/develop)

shaneray:
[...]

PR #444, opened 2017-02-11
Added DotSettings

Silvenga:
Per #443, adds `.DotSettings` based on `CONTRIBUTING.md`.

Basically default Resharper code styles + the following:
- Require explicit types
- Suggest var type on complex types
- Require braces

Everything else should be defaults. Settings should automatically be pulled into R#.


phil-scott-78:
I loaded this up with R# 2016.13.2 and I can confirm things look good in there.

daveaglick:
Thanks! I love this developer experience stuff but never seem to have enough time to focus on it. Appreciate the help.

PR #445, opened 2017-02-12
WIP: Adds helper functions to Pipeline for modifying module list

phil-scott-78:
Adds a handful of helpers for inserting new modules into an existing pipeline which should be very helpful for people needing to slightly tweak recipes. Goal is to be able to write code like this in the config.wyam file

```
Pipelines[BlogPipelines.Posts].InsertBeforeFirst<Razor>(new Highlight());
```

or 

```
Pipelines[BlogPipelines.Resources].ReplaceFirst<CopyFiles>(new CopyFiles("**/*{!.cshtml,!.md,!.scss,}"));
```

I had to stop myself from boiling the ocean with these methods though. A few scenarios that this won't cover would be

* Replacing the Title `Meta` module in the Tags pipeline. It's the second item in there. One thing that would help here is an overload that takes a predicate for additional filtering on each method, but that would double the size of everything. Wanted feedback before I spam that out.
* `Concat` modules. Jamming something into those isn't possible right now. If you had to, for example, tweak `ReadFiles("tags/index.cshtml")` in the Pages pipeline it would involve replacing the whole `Concat` module.

### DIFF
--- a/src/core/Wyam.Common/Modules/IModuleCollectionExtensions.cs
+++ b/src/core/Wyam.Common/Modules/IModuleCollectionExtensions.cs
@@ -1,0 +1,135 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Wyam.Common.Modules
+{
+    public static class ModuleCollectionExtensions
+    {
+        public static void InsertAfter(this IModuleCollection moduleCollection, string name, params IModule[] modules) 
+            => moduleCollection.Insert(moduleCollection.GuardIndexOf(name) + 1, modules);
+
+        public static void InsertBefore(this IModuleCollection moduleCollection, string name, params IModule[] modules) 
+            => moduleCollection.Insert(moduleCollection.GuardIndexOf(name), modules);
+
+        public static void InsertBeforeFirst<T>(this IModuleCollection moduleCollection, params IModule[] modules)
+            where T : class, IModule
+            => moduleCollection.InsertBeforeFirst<T>(_ => true, modules);
+
+        public static void InsertBeforeFirst<T>(this IModuleCollection moduleCollection, Predicate<T> filter, params IModule[] modules)
+            where T : class, IModule
+            => moduleCollection.Insert(moduleCollection.GuardIndexOfFirst(filter), modules);
+
+        public static void InsertAfterFirst<T>(this IModuleCollection moduleCollection, params IModule[] modules)
+            where T : class, IModule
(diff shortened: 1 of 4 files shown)


daveaglick:
This is excellent. Tempted to just merge it in, but here's some (hopefully quick) suggestions:
- I'm fine with the extra predicate methods.
- To avoid making the interface more complicated, all of these (existing + new predicate overloads) could be extension methods on `IPipeline` since it implements `IList<IModule>`. Even better, maybe make them extensions of `IList<IModule>` instead of `IPipeline` (see below).

---

Good point on `Concat`. To some extent, that'll be a limitation for any module that has submodules (there are several). If these methods are `IList<IModule>` extensions as mentioned above, and if we implemented `IList<IModule>` for modules that accepted submodules as appropriate, then those would light up with these extensions and we could do things like:
```
Pipelines[BlogPipelines.Pages].OfType<Concat>().Skip(2).First()
  .InsertAfterFirst<ReadFiles>(new Meta(...));
```

This also brings up a mostly unrelated point, but something I've thought about - it would be interesting to give modules names like we do for pipelines. I don't think it would be appropriate in all cases, and certainly not for custom configurations since the name only matters if you need to find a module later, but it might make editing existing recipe scenarios easier. I.e., being able to write this would be cool:
```
Pipelines[BlogPipelines.Pages].InsertBefore(BlogPipelines.PagesModules.TagsIndex, new Highlight());
```

I'll open a separate issue for this for another day...

daveaglick:
Okay, you've successfully [nerd sniped](https://xkcd.com/356/) me. I'm going to go ahead and finish off #447 and #448 while I'm thinking about them. That should open the door for you to finish this one off.

daveaglick:
Okay, #447 and #448 are now done. You can make the extensions work against `IModuleCollection` and that should get us support for both pipelines and modules that contain children.

phil-scott-78:
Nice. I'll pull those changes. My wife is working a night shift tomorrow night so making these an extension method will be my valentines date.

daveaglick:
Haha, so does that mean I owe you flowers 🌷?  Or would you settle for a beer if we ever meet in person?

phil-scott-78:
Roses are red.
Violets are blue.
Thanks for the `IModuleCollection`.
'Cause the extensions are in the appveyor queue.

phil-scott-78:
Everything is passing. Not thrilled with the syntax for modifying the modules, but the important thing is that it is actually possible to do. Which is quite nice.

daveaglick:
> Roses are red.

That was...beautiful :cry: 

Thanks, especially for the bug fixes - hadn't gotten around to unit tests for `ModuleCollection` yet, so you saved me from embarrassment (this time).

Issue #385, opened 2017-01-14
Move the preview server out into a separate library.


pauldotknopf:
I am referencing the core libraries, and building things manually. I'd also like to be able to use your preview server and owin middleware. Currently, the server code is embedded in ```Wyam.exe```.

daveaglick:
That's a great idea.

If it exists outside the console application, how would you start it? Should the preview server API support firing it up from some other consumer or should the new library also be a console application?

pauldotknopf:
I'd say just a library.

Also, we might want to think about moving from Owin, and to Kestrel. This will set us up better for .NET Core, when the port happens.

Maybe I could create a new library completely, using the new Kestrel stuff? I'd love to help out if I could.

I am fucking loving this project.

edit: For some context of what I am doing, I am building a software interface for medical device documentation. It is a beast to tackle, and the people doing the documentation are tech-illiterate. I am embeddeding Wyam into a peice of software to allow the documentation writers to not have to learn command prompt, but still have all the build/preview abilities.

daveaglick:
> I am fucking loving this project.

That's what I like to hear! Comments like this keep me going :smile:

The extra context was helpful. I agree - splitting out into a separate library with a nice public API is the way to go. Also agree it'll be a good opportunity to update to the latest Kestrel libraries.

Feel free to take a stab at it. I'll add it to my queue, but I'm planning on tackling performance (NuGet and otherwise) next so it may be a little while until I get to it. If you beat me to it, all the better.

daveaglick:
Mostly unrelated to this issue, but perhaps helpful for what you're working on, have you seen Netlify CMS? It's something I've been meaning to dig into, looks like a client-side generator-agnostic frontend for editing static site files. Might be relevant (or might not be) but figured I'd mention it.

https://github.com/netlify/netlify-cms

pauldotknopf:
That actually looks like exactly what I am building. I will look into it. Thanks!

daveaglick:
FYI - going to start working on this one in the next couple of days.

daveaglick:
Wyam.Hosting is now complete, it'll go up on NuGet with the next release. The nice thing is that it's not dependant on any other Wyam libraries at all. It can be consumed as a stand alone Kestrel-based server for static sites.

Note that right now it's limited to Windows and .NET 4.6.2. Most of the bits are in place to switch it to netstandard soon - just have to wait for a couple more external things to drop (like VS 2017 RTM).

PR #449, opened 2017-02-14
Moves Preview Server to Use Kestrel

Silvenga:
This pull request adds support for the Kestrel server.

> Kestrel is a cross-platform web server for ASP.NET Core based on libuv, a cross-platform asynchronous I/O library. Kestrel is the web server that is included by default in ASP.NET Core new project templates.

The move was easier than expected, most of the changes are the new packages, the preview server is virtually the same. 

A basic pro/cons list (some are IMHO):

## Gains

- Trivial support for HTTPS.
- Cross platform.
- No longer uses http.sys.
  - No more issues with url reservations under Windows (I have very strong feelings for `urlacl` 😒).
  - Not hostname verification.
- Kestrel will replace Katana.
- Easier support for dotNetCore.
- WebSockets! <- makes #420 a bit more possible.

## Losses

- No longer uses http.sys.
- No support for port sharing with IIS (did anyone use this anyway?)
- Complicated short term support:
  - Native nuget packages are not supported out-of-box < VS2017 projects (custom includes required). This is mitigated by a move to dotNetCore (planned?).
  - Only supports Windows x86 with the default build. This is also mitigated by either upgrading to VS2017 or dotNetCore.

## Considerations

- Large functional move with little benefit in the majority of cases. I'm primarily concerned with what is unknown with Kestrel - not much community usage yet + not much personal experience.
- Reduces portability for now, but Wyam doesn't support Mono anyway, so moot?

### DIFF
--- a/src/clients/Wyam/Wyam.csproj
+++ b/src/clients/Wyam/Wyam.csproj
@@ -268,7 +268,7 @@
     <Compile Include="Commands\ConfigOptions.cs" />
     <Compile Include="Commands\HelpCommand.cs" />
     <Compile Include="Commands\PreviewCommand.cs" />
-    <Compile Include="Owin\IApplicationBuilderExtensions.cs" />
+    <Compile Include="Server\ApplicationBuilderExtensions.cs" />
     <Compile Include="Owin\VirtualDirectoryExtensions.cs" />
     <Compile Include="Owin\VirtualDirectoryMiddleware.cs" />
     <Compile Include="PreviewServer.cs" />
@@ -296,7 +296,7 @@
     <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
-	<!-- HACK, required until VS2017 and the new project structure. -->
+    <!-- HACK, required until VS2017 and the new project structure. -->
     <Content Include="..\..\..\packages\Libuv.1.9.1\runtimes\win7-x86\native\libuv.dll">
       <Link>libuv.dll</Link>
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
--- a/src/clients/Wyam/Wyam.csproj
+++ b/src/clients/Wyam/Wyam.csproj
@@ -49,6 +49,84 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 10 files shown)


src/clients/Wyam/Wyam.csproj
```diff
@@ -139,6 +288,14 @@
...
   </ItemGroup>
   <ItemGroup>
+    <!-- HACK, required until VS2017 and the new project structure. -->
+    <Content Include="..\..\..\packages\Libuv.1.9.1\runtimes\win7-x86\native\libuv.dll">
+      <Link>libuv.dll</Link>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <Visible>False</Visible>
+    </Content>
+  </ItemGroup>
+  <ItemGroup>
```
Silvenga: This is the hacky include. This assumes a x86 compatible build.

src/clients/Wyam/Wyam.csproj
```diff
@@ -139,6 +288,14 @@
     <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
+    <!-- HACK, required until VS2017 and the new project structure. -->
```
Silvenga, line 291: This is the hacky include. This assumes a x86 compatible build.

daveaglick:
Totally on board with this. In fact, there's an open issue #385 to move the preview server to an entirely separate library so what I'll probably do is merge this in and then split it out. Then the work in #420 can move to the new `Wyam.Server` library.

Silvenga:
Sweet! 

I would propose a merge of this and #420 before the break out so that #420 doesn't need to be rebuilt. #420 essentially works, just not under Windows 7, worst case the changes are purely additive.

daveaglick:
That sounds like a plan - no reason to make you deal with the split while trying to work out the rebase. We'll merge this one as-is, take another look at LiveReload, hopefully be able to merge that one now, and then I'll split them out. That'll be a better verification anyway since we'll know it worked pre-split.

Hopefully I'll get a chance to merge this one later tonight or early tomorrow.

daveaglick:
Worked fine on my test merge, so I'll pull this in and we'll deal with any other problems as they arise.

PR #450, opened 2017-02-14
Updated LibGit2Sharp to 0.23.1.

pauldotknopf:
I am using LibGit2Sharp directly, and I required a newer version.


pauldotknopf:
Also, I checked the existing git unit tests, and they were failing. So, I couldn't get the tests to pass with the new LibGit2Sharp version.

However, this does work locally with my recipe.

daveaglick:
Great! I'll pull it in even with the failing tests - I've had them disabled for a while because they take so long to run, and IIRC they look at the repo for certain files. I've refactored so many times it's entirely possible they're looking for files that aren't there anymore.

PR #453, opened 2017-02-16
Filters JetBrains assembly from loaded namespace list while profiling

phil-scott-78:
While looking at perf I ran into a weird issue where DotTrace caused the configuration and razor modules to blow up. Turns out that DotTrace shoves it's profiling assembly in dynamically when running a profile of the application. Not a huge deal but it gets picked up as a loaded namespace but the tools that write the configuration file using statements and with the razor parser. If fails to find that DLL and fails. 

This simply filters it out if found from the config.


daveaglick:
I just YOLO merged this

PR #457, opened 2017-02-22
Upgrade Cake to 0.17.0

### LINKED ISSUE #437: Update to latest Cake
With the upcoming release of 0.18.0, there is a requirement that all addin's have a minimum version of Cake, i.e. 0.16.2 and above.

@daveaglick would you be in a position to update this addin to the latest Cake version, so that it will be supported OOTB?

Thanks!

pascalberger:
Cake 0.18.0 will requires at least version 0.16.x to work.

Fixes #437


daveaglick:
Awesome, thanks! One more thing I don't have to think about :smile:

PR #463, opened 2017-03-05
If comments are repeated, we only need it to display once

### LINKED ISSUE #460: De-dupe doc comments for partial classes
If a partial class has the same doc comment in more than one place, it shows up multiple times in the summary. Should only show distinct doc comments (but still aggregate from multiple partials).

M-Zuber:
Fixes #460


daveaglick:
Awesome, thanks!

PR #465, opened 2017-03-07
Resolves 500 Error w/ LR Injection

Silvenga:
Found a bug in the LiveReload injection middleware, I've added a test case to show it. It presents as 500 errors on any HTML page that does not have a ending body tag.

The variable `interceptedBody` is closed by the `using` statement on `reader` resulting in an `ObjectDisposedException`. Removed usage of the `using` statement to prevent early closure (recommend for IDisposables, not strictly required for MemoryStreams).

## Considerations

Sooo, it turns out the HTML minifier included with Wyam actually correctly removes the ending `html`/`body` tags (took me a while to figure it out). According to (https://www.w3.org/TR/2011/WD-html5-20110525/syntax.html#optional-tags)

> A body element's end tag may be omitted if the body element is not immediately followed by a comment.

### DIFF
--- a/src/core/Wyam.Hosting/Owin/ScriptInjectionMiddleware.cs
+++ b/src/core/Wyam.Hosting/Owin/ScriptInjectionMiddleware.cs
@@ -2,11 +2,9 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


daveaglick:
Ah, that makes total sense now - great catch.

daveaglick:
I'm hoping to get a patch release out today with a few more roll-up fixes, so this should go out very soon.

PR #466, opened 2017-03-08
WIP: Caching Module(s)

### LINKED ISSUE #346: RestoreFiles module and caching between executions
Build a module that will hash the content of input documents and then store that hash along with a specified key from the document (document source, user-defined metadata key). If the hash matches the previous one, don't pass the document as output. If it doesn't continue processing the document. For example, given an image with a time consuming compression phase don't recompress it unless the image actually changes.

Need someplace to put the set of keys and hashes - perhaps make this a specified file path but from the build root path (not output path)

If the hash matches and we stop processing, how do we make sure the previously generated result(s) still make it to output? We clear the output folder at beginning of generation so it can't come from there. Also, the dropped document could have resulted in multiple output files - how to know? Might need to think about this a bit...

Silvenga:
This pull request attempts to resolve #346 and focuses primarily on quality of life speed improvements when using the preview server.

## Considerations

Wyam's greatest strength (IMHO) is its flexibility with its pipelines. This, however, makes caching an incredibly difficult problem to solve. To mitigate this somewhat I propose targeted caching modules, rather than attempting to compensate for every edge case.

## Module(s) (pending a better name)

- [ ] SerialCache - This module takes in a stream of inputs and considers each individually for caching. These inputs may output one or many outputs which are cached under the input. When the input is seen again without modifications, the outputs are simply returned. Currently CRC32 hashing is occurring on the document content with the cache being stored on the disk. This caching module assumes that each input can be handled independently without side effect.

```
engine.Pipelines.Add(PipelineKeys.Pages,
    new ReadFiles(ctx => "*.md"),
    new SerialCache(
        new FrontMatter(new Yaml.Yaml()),
        new Execute(ctx => new Markdown.Markdown()),
        new Razor.Razor(),
        new MinifyHtml()
    ),
    new WriteFiles()
);
```

## Notes

- I'm currently seeing 3x speed improvement on my blog.
- I will fix Unit Test naming.
- Still a lot of work to do, but want to open for feedback early.
- First module, hope I'm not doing something blatantly bad. 😸

### DIFF
--- a/src/core/Wyam.Core/Modules/Control/SerialCache.cs
+++ b/src/core/Wyam.Core/Modules/Control/SerialCache.cs
@@ -1,0 +1,125 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+
+using Wyam.Common.Documents;
+using Wyam.Common.Execution;
+using Wyam.Common.IO;
+using Wyam.Common.Meta;
+using Wyam.Common.Modules;
+
+namespace Wyam.Core.Modules.Control
+{
+    /// <summary>
+    /// TODO Write this
+    /// </summary>
+    public class SerialCache : ContainerModule
+    {
+        // Considerations
+        // - This module assumes a single input/multiple outputs
+        // - Should we cache selectively?
+        // TO.DO
+        // Consider cacheability based on metadata and source path, not just content.
+
+        // TODO Move this to key object
+        private const string InputHashKey = "SerialCache.InputHash";
+        private const string FromCacheKey = "SerialCache.FromCache";
+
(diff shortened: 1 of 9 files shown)


LokiMidgard:
If I see it correctly, the 'sub modules' get executed once for each document. You should highlight this in the documentaiton. Some modules require that the complete list of documents will be processed at once.

Silvenga:
@LokiMidgard correct. I'm also thinking we should have a module that can take a list of documents to be processed in entirety in which the output is cached based on the combination of inputs (supports many to one relationships). This module would of course be significantly less efficient as it cannot selectively regenerate based on inputs.

Silvenga:
Going to close this as I don't have much time right now, I'll work on this when I can find more time.

daveaglick:
@Silvenga Totally understand the time limitation. That said, I really liked where you were going with this (even if I didn't have enough time myself for a formal review). I'd love to see it resurrected at some point in one form or another. At a minimum, I'll come back here if I ever start working on #346 myself.

PR #479, opened 2017-03-18
Image compression

schulz3000:
implement #323
Compress images lossy or lossless using some external image compression
tools.
Integrated in Wyam Image module.
It's working for jpg, png and gif (also animated ones)
Depend on input image count and size it can take some seconds to minutes
to process.

### DIFF
--- a/src/extensions/Wyam.Images/Compressor.cs
+++ b/src/extensions/Wyam.Images/Compressor.cs
@@ -1,0 +1,106 @@
+﻿using System.Diagnostics;
+using System.Drawing.Imaging;
+using System.IO;
+using System.Linq;
+
+namespace Wyam.Images
+{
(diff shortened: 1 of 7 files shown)


daveaglick:
Thanks for your efforts here! Unfortunately, I don't think I'm going to be able to pull this in. I've tried really hard to stay away from native libraries or "shelling out". Both approaches introduce a number of complexities regarding platform dependence, bitness, and distribution. This will only get magnified once we make a planned move to cross-platform .NET Core in a few months.

If there are .NET libraries to perform this kind of compression I'd be happy to take another look at this functionality with those.

schulz3000:
understand. I didn't thought about cross platt.
I will try to review and maybe make an additional PR for it in the future.

PR #509, opened 2017-05-05
Reuse the same ASP.net objects to avoid compiling the same pages agai…

jontdelorme:
Hello, we are using Wyam internally to generate our websites.

Recently I added a blog module and the hot reload time went up to more than 30 seconds for a relatively small web site. I found it reduced a lot my appreciation of the product in comparison to my 5 seconds previewing time I had before.

After some profiling, I found that it was spending most of its time in Razor which was recompiling the same cshtml files over and over.

In this merge request, I removed the states to a maximum so I can reuse the same ASP.net objects on hot reloads. With that, I have a time of around 1 second for a live preview instead of around 40 otherwise.

The only thing that I'm not really comfortable with is that these ASP objects do not seem thread safe so I had to use a mutex. I'm sure there is a better way, I just don't know that much Razor concepts and where the cache lives.

### DIFF
--- a/src/extensions/Wyam.Razor/DynamicAssemblyCollection.cs
+++ b/src/extensions/Wyam.Razor/DynamicAssemblyCollection.cs
@@ -1,0 +1,38 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Wyam.Razor
+{
+    internal class DynamicAssemblyCollection : IEnumerable<byte[]>
+    {
+        private readonly IReadOnlyCollection<byte[]> _assemblies;
+
+        public DynamicAssemblyCollection(IEnumerable<byte[]> assemblies)
+        {
+            _assemblies = assemblies.ToArray();
+        }
+
+        public IEnumerator<byte[]> GetEnumerator()
+        {
+            IEnumerable<byte[]> assemblies = _assemblies;
+            return assemblies.GetEnumerator();
+        }
+
+        public override int GetHashCode()
+        {
+            return _assemblies.Count;
+        }
+
+        public override bool Equals(object obj)
+        {
+            var other = obj as DynamicAssemblyCollection;
+            return other != null && _assemblies.SequenceEqual(other._assemblies);
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+    }
+}
--- a/src/extensions/Wyam.Razor/HostingEnvironment.cs
+++ b/src/extensions/Wyam.Razor/HostingEnvironment.cs
@@ -1,21 +1,22 @@
 ﻿using System.Reflection;
 using Microsoft.AspNetCore.Hosting;
-using Microsoft.Extensions.FileProviders;
 using Wyam.Common.Execution;
+using Wyam.Common.IO;
+using IFileProvider = Microsoft.Extensions.FileProviders.IFileProvider;
 
 namespace Wyam.Razor
 {
     internal class HostingEnvironment : IHostingEnvironment
     {
-        public HostingEnvironment(IExecutionContext context)
+        public HostingEnvironment(IReadOnlyFileSystem fileSystem)
         {
             EnvironmentName = "Wyam";
 
             // This gets used to load dependencies and is passed to Assembly.Load()
             ApplicationName = typeof(HostingEnvironment).Assembly.FullName;
 
-            WebRootPath = context.FileSystem.RootPath.FullPath;
-            WebRootFileProvider = new FileSystemFileProvider(context.FileSystem);
+            WebRootPath = fileSystem.RootPath.FullPath;
+            WebRootFileProvider = new FileSystemFileProvider(fileSystem);
             ContentRootPath = WebRootPath;
             ContentRootFileProvider = WebRootFileProvider;
         }
--- a/src/extensions/Wyam.Razor/MetadataReferenceFeatureProvider.cs
+++ b/src/extensions/Wyam.Razor/MetadataReferenceFeatureProvider.cs
@@ -5,17 +5,16 @@ using System.Reflection;
 using Microsoft.AspNetCore.Mvc.ApplicationParts;
 using Microsoft.AspNetCore.Mvc.Razor.Compilation;
 using Microsoft.CodeAnalysis;
-using Wyam.Common.Execution;
 
 namespace Wyam.Razor
 {
     internal class MetadataReferenceFeatureProvider : IApplicationFeatureProvider<MetadataReferenceFeature>
     {
-        private readonly IExecutionContext _executionContext;
+        private readonly DynamicAssemblyCollection _dynamicAssemblies;
 
-        public MetadataReferenceFeatureProvider(IExecutionContext executionContext)
+        public MetadataReferenceFeatureProvider(DynamicAssemblyCollection dynamicAssemblies)
         {
-            _executionContext = executionContext;
+            _dynamicAssemblies = dynamicAssemblies;
         }
 
         public void PopulateFeature(IEnumerable<ApplicationPart> parts, MetadataReferenceFeature feature)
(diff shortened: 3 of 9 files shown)


daveaglick:
That's awesome, thanks! Every bit of performance tuning helps, especially since I'm focused mainly on features and architecture at this point. I'm just about to do a release and then I'll circle back and review this for the next version.

Daniel15:
I'm confused as to the purpose of the custom collection classes (`DynamicAssemblyCollection` and `NamespaceCollection`) - Why not use generic collections? `NamespaceCollection` seems to just be a hand-rolled version of `ReadOnlyCollection<T>`.

daveaglick:
Finally getting around to reviewing this PR, thanks for your patience! It looks good - a couple comments:
* A mutex for locking is fine. There were some issues in the ASP.NET repository where thread-safety for Razor was brought up and team basically confirmed that part of the code isn't thread safe if reusing services.
* The one problem I see is with using the last write time as a cache key, and specifically the use of `System.IO.FileSystem`. Wyam uses a file system abstraction and it's not guaranteed that files actually live in the disk file system (most do now, but that may not always be true). At a minimum, we should get the write time from the abstraction (if it were exposed, which I don't think it is right now). Beyond that, write time isn't a great determinator anyway because documents could have been manipulated as part of the pipeline processing even if the underlying file didn't change. Using a hash of the document content is probably a better cache key - I'll investigate how feasible that is and make the changes after merge if so.

daveaglick:
@Daniel15 Believe I'm in a position to answer your question about `DynamicAssemblyCollection` and `NamespaceCollection` now that I've spent some time in this PR. My initial impression was the same as yours, but it turns out there are two reasons separate classes are needed. The first is that they are included as part of a cache key object that uses the default equality comparer. That means that special behavior is needed if an item-by-item comparison is desired, which it is. The second reason is that these collections are registered with the Razor service container as singletons which means they need to be uniquely typed in order to be properly injected into the constructors of other services.

Anyway, thought I'd leave the reasoning up here for posterity in case it comes up later.

daveaglick:
@jontdelorme Thanks again for this great PR!

I was almost ready to merge this - I had converted the cache key for document-based views to use a hash of the content instead of the last write time as discussed above and I was seeing 10% - 15% improvements when cold rending (probably due to fewer service instantiations) and 20% - 30% improvements on hot rebuilds for some blogs and other sample sites. It was awesome!

But then I ran the changes against a large docs build and the generation time got much, much worse. It appears there's a good improvement for small numbers of documents, but at some point as more documents are rendered it actually starts to hurt. I'm guessing this may have something to do with generating a hash code for the cache key - it means we have to stream each document twice since we need the cache key before streaming it to the Razor compiler. Over hundreds or thousands of documents, I guess that adds up.

I'm going to think about how best to handle this so we can still get the advantages of this PR without killing performance on large generations. Some idea I have are pre-generating the document hash when it's created (may be valuable to other modules as well) or using the full content as the cache key instead of a hash.

jontdelorme:
@daveaglick thanks for taking time to review this PR!

You were right about the issue with the timestamp. I am now also using a hash of the content instead. Should I update my merge request with it or you prefer to use your own fix?

In my profilings, the best gains were more from the reuse of razor context than from the manual "compilation cache". It looks like razor keeps an internal cache and it makes a difference for files that are always rendered like layout.cshtml. Maybe you can try to deactivate the compilation cache to see if you still benefit from the first gains.

I can also have a look. Which web site should I use to get the same results? By curiosity, how long does it take on your side to get a "live preview" of your changes?

daveaglick:
I'm currently seeing gains in everything but the Wyam docs site, which is running much slower. If you want to play with it, it's in the Wyam.Web repo in this same GitHub org. You'll need to run a normal build with the "Preview" target on the current release to download the source and set everything up, but then can use the "Debug" target to generate using the local Wyam build. Check out the Cake file for the details.

That site has nearly 3,000 pages for the API, so it's a great example of a "large" site. After a bit of profiling, I'm pretty sure the problem is the mutex and contention. Thankfully that should be pretty easy to clear up - I think replacing it with a new service context might do the trick.

I've gotten pretty far (and made some other somewhat unrelated changes) in a branch. It's probably diverged a bit from the original PR so I'll commit it to the repo tomorrow so you can take a look and let me know what you think. I think we're getting close - looking forward to getting this into the next release!

jontdelorme:
Okay, I'll synchronize with your changes then!

It makes sense that if you are rendering a lot of different pages generated on the fly, the mutex kills the optimization. 

I haven't started to generate the api doc on my side, that explains why I didnt experiment the slow down of single threading.

daveaglick:
@jontdelorme I pushed my changes to the `pr509` branch (https://github.com/Wyamio/Wyam/tree/pr509) - take a look and see what you think. Some were just style changes. Here's a (mostly) full set of what I did:
* Changed the compiler cache key to not use the file path and use a hash instead of the write date
* Removed the locking code and replaced with a service scope
* Split out the internal classes
* Renamed the `RazorInterpreter` to `RazorCompiler` for consistency
* Added an execution ID to the Wyam execution context and used it to clear the internal Razor cache for partials/layouts on new generations (this isn't directly related to your changes, but I discovered it while working this PR)

I'm very curious if it still works as well for you after my changes. I'm now seeing marginal improvements for the big Wyam docs site and great improvements for smaller sites. If your internal use case is also still improved, I'll merge everything in.

jairbubbles:
@jontdelorme The cache key is only for preview mode right? (there is no persistency) Why not using a file watcher to detect when a file is modified and invalidate the cache for this file?

jontdelorme:
Ok awesome. I will try it tomorrow

Le 13 juin 2017 à 12:32, jairbubbles <notifications@github.com<mailto:notifications@github.com>> a écrit :


@jontdelorme<https://github.com/jontdelorme> The cache key is only for preview mode right? (there is no persistency) Why not using a file watcher to detect when a file is modified and invalidate the cache for this file?

daveaglick:
@jairbubbles @jontdelorme A watcher won't work for the same reason the file write time wouldn't work. Even though we're only using the local file system right now, modules need to be built only accessing file data through the `IFileSystem` abstraction. That will allow alternate file systems in the future (such as unextracted zip files, straight from GitHub, FTP, etc.)

Also note that the content of a *document* isn't the same as the content of a *file* - the document content could get mutated along the way based on changes to a totally different file depending on the pipelines and modules in use so using a watcher to invalidate the cache for a given document based on underlying files is also unreliable.

It's been in my long-term plans to add change watching to the file system abstraction for all file systems that will support it (and probably do some sort of polling for those that don't), but that's a ways off. This is especially complicated for the document vs. file reason above because we'll have to "track" all the contributing files to a given document and trigger for any of them. In the meantime we need to be careful to steer clear of anything that's directly tied to the local file system. The hashing appears to be pretty performant anyway - it's just a MD5 constructed straight from the stream, and in my profiling it doesn't seem to be the part that's taking cycles.

jairbubbles:
Makes totally sense. Isn't there any existing lib out there which makes such an abstraction? If not, it would make sense to externalize this as it would be useful for pretty much all applications (just to be able to mock properly for unit testing for instance).

EDIT: seems there is one in .NET Core (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/file-providers)

daveaglick:
@jairbubbles Yeah, great question about externalizing. I've thought about it, and it probably would make a nice stand-alone library. Main thing holding me back at this point is time - I just don't have enough to maintain another OSS project right now. And some parts like the globbing support will be a little challenging to extract.

@xoofx actually has a nice independent virtual file system at https://github.com/xoofx/zio. It was built much later than the one in Wyam, but has a lot of the same goals and features (fun fact - it's also for a static site generator!)

jairbubbles:
@daveaglick The one from .NET core has a `Watch` method. I guess we coders have all the same dreams :smile: 
(NuGet package: https://www.nuget.org/packages/Microsoft.Extensions.FileProviders.Abstractions)

jairbubbles:
@daveaglick I've already been on Zio depot but I totally forgot about it! The interface seems to be way too complicated for your user cases (GitHub, FTP, zip etc...) though: [Zio doc](https://github.com/xoofx/zio/blob/master/doc)

jontdelorme:
@daveaglick, I just tested your improvements over my PR and it works great. I get similar results to what I had before and it's pretty cool being able to remove the lock, 

Thanks for your time!

daveaglick:
Awesome!! I'll plan to merge this tomorrow and hopefully get another release out soon.

daveaglick:

daveaglick:
All merged in, thanks again!!

PR #529, opened 2017-06-07
suppress CS1701 and related diagnostics instead of filtering it out

filipw:
It doesn't really fix anything, but is probably a slightly more elegant API usage.


daveaglick:
You giving Wyam a spin or just doing some code review? In either case, I'm all for more elegant code - thanks a bunch!

filipw:
both 😀

PR #535, opened 2017-06-17
docs: OutputDirectory is incorrect, should be OutputPath

ghuntley:


daveaglick:
👍 So much API has changed in the last year, finding these incorrect examples is great - thanks.

daveaglick:
Thanks!

Issue #542, opened 2017-07-19
New version number breaks AppVeyor build


miroslavpopovic:
I have recently migrated my blog to Wyam. One of the things I set up was AppVeyor build according to official documentation - https://wyam.io/docs/deployment/appveyor.

The last commit to Wyam's `master` branch modified the `RELEASE` document to `v0.18.5`. AppVeyor build script reads from that file to figure out the latest version and then downloads the zip file from releases. At this moment, there's no `v0.18.5` version in releases which causes AppVeyor build to fail.

@daveaglick when can we expect the new release to be available for download, and more importantly how could we avoid this problem in future? 

I guess creating a release is a manual operation and the `RELEASE` document in `master` branch can be out of sync with release downloads. Not sure what else could be done... maybe querying GitHub API for the latest release...

daveaglick:
Working on it right now. Normally the master branch updates concurrent to a release (with a short 1 or 2 minute window for uploading binaries to NuGet and GitHub, NuGet cache refreshes, etc.). The [NuGet problem AppVeyor is experiencing](http://status.appveyor.com/incidents/m2vdvw39kdk8) unexpectedly gummed everything up and delayed the release since I do multiple CI build and test steps before finally releasing. Hoping to get the actual release out in the open within the hour - if I can't I'll roll back the version in master until I figure out how to work around the CI problems.

daveaglick:
0.18.5 is up on both NuGet and the GitHub releases page. The NuGet cache may take a few minutes to update, but the GitHub binaries are already available. That should resolve any CI problems you were having - sorry!

PR #543, opened 2017-07-19
Latest version is still v0.18.4

Pothulapati:
The latest downloadable version from the releases page is still **v0.18.4**.  It is breaking my continuous deployment process. I can move on by not getting the latest version from this file every time but I've seen people doing the same.
So,I thought it would be better if I bring this issue to notice


daveaglick:
See #542 - the actual 0.18.5 should be going out soon. There was an unexpected problem with the release due to a AppVeyor/NuGet routing issue.

daveaglick:
I'm going to close this PR - I'll manually roll back the version number in master if the release doesn't make it out in the next few minutes.

PR #545, opened 2017-07-24
Upgrade to latest JSPool

Daniel15:
Upgrades Wyam to the latest version of JSPool. 

One of the main changes in this release is that engines are automatically returned to the pool when they are disposed. This means that Wyam's `JsEnginePool` and `PooledJsEngine` wrappers are no longer required since JSPool has a very similar interface now. I deleted these wrappers, and also removed some of the unnecessary layers of abstraction in order to simplify the code.


src/core/Wyam.Testing/Execution/TestExecutionContext.cs
```diff
@@ -3,15 +3,18 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using JavaScriptEngineSwitcher.Core;
+using JSPool;
+using NSubstitute;
```
Daniel15, line 11: Oops, I forgot to remove this. It's not being used

src/core/Wyam.Testing/Execution/TestExecutionContext.cs
```diff
@@ -237,26 +242,43 @@ public TestJsEnginePool(Func<IJsEngine> engineFunc, Action<IJsEngine> initialize
...
+            public PooledJsEngine GetEngine(TimeSpan? timeout = null)
             {
-                IJsEngine engine = _engineFunc();
-                _initializer?.Invoke(engine);
-                return engine;
+                // PooledJsEngine's InnerEngine setter is internal, so we need to do some reflection acrobatics for now.
+                // TODO: https://github.com/Daniel15/JSPool/issues/24
+                PooledJsEngine pooledEngine = new PooledJsEngine();
+                Type engineType = pooledEngine.GetType();
+                engineType.GetProperty("InnerEngine").SetValue(pooledEngine, _engineFunc());
```
Daniel15, line 251: I'll fix this in JSPool, but for now we need this ugly hack :(

tests/extensions/Wyam.Highlight.Tests/HighlightFixture.cs
```diff
@@ -208,17 +207,19 @@ public void HighlightFailsForMissingLanguage()
...
 
                 // When, Then
-                Assert.Throws<AggregateException>(() =>
+                AggregateException ex = Assert.Throws<AggregateException>(() =>
                 {
                     List<IDocument> results = highlight.Execute(new[] { document }, context).ToList();
                     Assert.IsNull(results, "Should never get here due to exception");
                 });
+                Assert.AreEqual(1, ex.InnerExceptions.Count);
+                StringAssert.Contains("Unknown language: \"zorg\"", ex.InnerExceptions[0].Message);
```
Daniel15, line 222: This test was only asserting that an AggregateException was thrown, but it wasn't actually checking the message of the assertion! I added some extra assertions to ensure the right exception is being caught.

Daniel15:
Hmm, the build works on my computer but on AppVeyor it's failing with:

> System.IO.FileLoadException : Could not load file or assembly 'JavaScriptEngineSwitcher.Core, Version=2.4.9.0, Culture=neutral, PublicKeyToken=c608b2a8cc9e4472' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)

 I suspect I'm just missing an assembly binding reference somewhere, or forgot to upgrade a package somewhere.

Daniel15:
@daveaglick Do you think you'll pull something like this into Statiq? Seems like Statiq.Framework is still on the old JSPool version.

daveaglick:
Yes, absolutely. I wasn’t actually intending for all the PRs to close, but it makes sense now that I think about it since I removed the old branches 😬. It’ll be worth it in the end, but all this shuffling is confusing even me.

I’ll go back through this and the other pending PRs to see if anything can be pulled into Statiq Framework. Feel free to open a new one there if you beat me to it (or even just an issue so we don’t forget).

PR #547, opened 2017-07-29
[wip] add wyam.youtube

ghuntley:
Greetings @daveaglick .

Opening this early as this is the first time playing with the internals of Wyam. Definitely not ready for merging yet but the general idea here is to provide a way to query Youtube from Wyam. My use case is to retrieve a list of videos in a playlist and then for each video bind the model to a razor partial view. 

I'm using the official Google library for interacting with YouTube and because there's just so many things one can actually do with this library I'm thinking the best way to handle it to follow the design of the GitHub module - i.e. don't define an abstraction.

Appreciate your thoughts on designing of the public API surface. 

Refer to https://github.com/reactiveui/website/blob/master/src/reactiveui.net/Services/YouTubeShowsService.cs#L74 for insights on how I'm currently using `Google.Apis.YouTube.v3`

Regards,
Geoff


src/extensions/Wyam.YouTube/YouTube.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            _requests[key] = request;
+            return this;
+        }
+
+        /// <inheritdoc />
+        public IEnumerable<IDocument> Execute(IReadOnlyList<IDocument> inputs, IExecutionContext context)
+        {
+            // YouTubeService is IDisposable.... how do you want this handled?
```
ghuntley, line 91: Here's the upstream library implementation:

* https://github.com/google/google-api-dotnet-client/blob/master/Src/Support/Google.Apis/Services/BaseClientService.cs#L134
* https://github.com/google/google-api-dotnet-client/blob/master/Src/Support/Google.Apis/Services/IClientService.cs#L35

ghuntley:
Refer to https://github.com/reactiveui/website/blob/master/src/reactiveui.net/Services/YouTubeShowsService.cs#L74 for insights on how I'm currently using `Google.Apis.YouTube.v3`

gep13:
@ghuntley ooo, this sounds very interesting! Is the intention here to create a playlist of say tutorial videos, and then display them as a page onto your Wyam site?

src/extensions/Wyam.YouTube/YouTube.cs
daveaglick: Hmm, that's a good question. Normally I'd say just create and dispose the service instance within the execute method since the documents typically only carry the result metadata with them. But in this case since you're modeling it after the GitHub module and exposing the service as a leaky abstraction, the question of how to create and dispose the service is trickier. I might need to make a callback or event or something in the execution context to let modules clean up service classes like this when the full execution is done (documents and their metadata could be used any time up until then). Let me think on it and get back to you - for now just don't worry about disposal and we'll circle back.

ghuntley:
> @ghuntley ooo, this sounds very interesting! Is the intention here to create a playlist of say tutorial videos, and then display them as a page onto your Wyam site?

Yup that's the intention.

gep13:
@ghuntley nice! I will be keeping a close eye on this issue, and your implementation, as I will have a use case for this as well 😄

src/extensions/Wyam.YouTube/YouTube.cs
daveaglick: So here's what I ended up doing - in the most recent commit to develop, any module that implements `IDisposable` will get disposed when the engine is disposed (and recursively through child modules if the module also implements `IEnumerable<IModule>`). That way you can instantiate a long-running service like this in the constructor and then clean it up when the engine is totally finished.

daveaglick:
Is this still WIP or is it ready for merge? If there are any small issues, I don't mind cleaning those up on merge. Doing a 1.0 push and would love to get this in.

ghuntley:
@daveaglick this is what I would love to pair with you on. There's stuff that I don't understand yet and my human debugging process w/wyam is really slow. I'm sure you know some tricks. Let's get those documented. Not ready for merge but you are 💯 okay to finish it for me.

daveaglick:
I would still really like to pair with you and @gep13 soon, but I want to get 1.0 out the door more :smile:. Blocking out video/chat time is hard for me, so I'm going to go ahead and finish this one off so I can get the release out the door. It looks mostly complete anyway, so probably just some minor touch-ups.

Then we can set a time and go over the changes so you get a better idea of why I made them.

gep13:
@daveaglick sounds like a good idea to me!

Hope the release goes well!

daveaglick:
@ghuntley This was in great shape - I basically didn't change anything other than moving the `YouTubeService` to an instance field so it could be disposed with the module. Thanks!

daveaglick:

Issue #97, opened 2015-10-27
Add flag to cache config assembly


daveaglick:
In console app, write generated config assembly to disk (by default?). Store hash of config file and regen if different. Need to figure out how or if this should be supported in embedded scenario.

Silvenga:
@daveaglick On my blog, config generation is taking around 4 seconds to run - would love to optimize that. I can write a pull for this - it looks straightforward:

- Mirror the implementation of `ConfigOptions.OutputScript`.
- When requested (or when not opted out), write the emitted config code to a dll, and store hash of config file.
- When requested and hashes match, use the compiled dll rather than compiling a new.

Hardest part is where to store the dll and hash (preferably in a self-cleaning way) and if this should be opt-in or opt-out. What are your thoughts?

File storage:
* User temp folder.
  * Harder to keep clean.
  * Needs considerations to prevent messing with other unknown instances of Wyam.
  * Keeps the files out of sight of the user.
* Hidden folder within the source directory (Like VS2017). 
   * Very easy to keep clean.
   * User has to deal with it (which might be a good thing).
   * Scoped to a single project, won't mess with others.

Opt'ing:
* I would be propose an opt-out flag unless hashing isn't capable of detecting if a regen needs to occur.

What do you mean by "supported in embedded scenario"?

daveaglick:
@Silvenga Your plan sounds perfect and would love a PR for this.

> where to store the dll and hash (preferably in a self-cleaning way)

I would prefer them to be somewhere in the source directory. Perhaps right alongside the config file. I.e., if the config file is `config.wyam` then generated assembly would be `config.wyam.dll` and hash could be `config.wyam.hash`. That way it's obvious which generated artifacts go with which config file in case there are more than one (which there might be in complex sites). Also allows moving them all as a unit to another source folder if you want to, since there's a 1-to-1 between config file and generated assembly that should be totally valid.

> if this should be opt-in or opt-out

I'm fine with it being opt-out. There's little chance of this causing problems since we're just writing the same thing we would be generating in memory anyway (as long as the hashing works), and it's a performance boost. Can't imagine there are enough downsides that anyone would choose not to cache the assembly except for niche scenarios.

We'll need to make a note somewhere about not checking the `.dll` and `.hash` (or whatever/wherever the files are) into source control. There's already another artifact that shouldn't be checked in (the `packages.xml` file) so it's probably time I make a special page in the docs about what to add to your `.gitignore`.

[...]

PR #557, opened 2017-08-23
Cache Emitted Config

Silvenga:
Implementation of #97. This pull adds the following.

* Added the ability to cache the `config.wyam` config generation.
* The cache will be broken when the MD5 hash of the code generated by the `DirectiveParser` differs in-between runs.
* Added the command directive `ignore-config-hash`, which forces evaluation even when the two hashes match (current and cached hash).

### DIFF
--- a/src/clients/Wyam/Commands/BuildCommand.cs
+++ b/src/clients/Wyam/Commands/BuildCommand.cs
@@ -69,7 +69,8 @@ namespace Wyam.Commands
             syntax.DefineOption("use-global-sources", ref _configOptions.UseGlobalSources, "Toggles the use of the global NuGet sources (default is false).");
             syntax.DefineOption("packages-path", ref _configOptions.PackagesPath, DirectoryPathFromArg, "The packages path to use (only if use-local is true).");
             syntax.DefineOption("output-script", ref _configOptions.OutputScript, "Outputs the config script after it's been processed for further debugging.");
-            syntax.DefineOption("verify-config", ref _verifyConfig, false, "Compile the configuration but do not execute.");
+            syntax.DefineOption("ignore-config-hash", ref _configOptions.IgnoreConfigHash, "Force evaluating the configuration file, even when no changes were detected.");
+            syntax.DefineOption("verify-config", ref _verifyConfig, false, "Compile the configuration but do not execute. Consider using the --ignore-config-hash directive when using this option.");
             syntax.DefineOption("noclean", ref _configOptions.NoClean, "Prevents cleaning of the output path on each execution.");
             syntax.DefineOption("nocache", ref _configOptions.NoCache, "Prevents caching information during execution (less memory usage but slower execution).");
 
--- a/src/clients/Wyam/Commands/BuildCommand.cs
+++ b/src/clients/Wyam/Commands/BuildCommand.cs
@@ -69,6 +69,7 @@ namespace Wyam.Commands
             syntax.DefineOption("use-global-sources", ref _configOptions.UseGlobalSources, "Toggles the use of the global NuGet sources (default is false).");
             syntax.DefineOption("packages-path", ref _configOptions.PackagesPath, DirectoryPathFromArg, "The packages path to use (only if use-local is true).");
 
+            syntax.DefineOption("no-output-config-assembly", ref _configOptions.NoOutputConfigAssembly, "Disable caching configuration file compulation.");
             syntax.DefineOption("ignore-config-hash", ref _configOptions.IgnoreConfigHash, "Force evaluating the configuration file, even when no changes were detected.");
             syntax.DefineOption("output-script", ref _configOptions.OutputScript, "Outputs the config script after it's been processed for further debugging. The directive --ignore-config-hash is required when using this option.");
             syntax.DefineOption("verify-config", ref _verifyConfig, false, "Compile the configuration but do not execute. The directive --ignore-config-hash is required when using this option.");
--- a/src/clients/Wyam/Commands/BuildCommand.cs
+++ b/src/clients/Wyam/Commands/BuildCommand.cs
@@ -68,9 +68,20 @@ namespace Wyam.Commands
             syntax.DefineOption("use-local-packages", ref _configOptions.UseLocalPackages, "Toggles the use of a local NuGet packages folder.");
             syntax.DefineOption("use-global-sources", ref _configOptions.UseGlobalSources, "Toggles the use of the global NuGet sources (default is false).");
             syntax.DefineOption("packages-path", ref _configOptions.PackagesPath, DirectoryPathFromArg, "The packages path to use (only if use-local is true).");
-            syntax.DefineOption("output-script", ref _configOptions.OutputScript, "Outputs the config script after it's been processed for further debugging. Consider using the --ignore-config-hash directive when using this option.");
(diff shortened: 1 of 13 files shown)


Silvenga:
@daveaglick I think this feature is complete, do you have any suggestions or concerns?

daveaglick:
AWESOME!! This is going to be great to get in for the 1.0 release soon. I'll review and merge as soon as I get a chance.

src/core/Wyam.Configuration/ConfigScript/CacheManager.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.IO;
+using System.Security.Cryptography;
+using System.Text;
+
+using Wyam.Common.Execution;
+using Wyam.Common.IO;
+
+namespace Wyam.Configuration.ConfigScript
+{
+    public class CacheManager
```
daveaglick: I like putting all the caching logic in a separate class, nice refactoring. Given that `CacheManager` is only instantiated in a method scope and isn't stored as a member anywhere, does it make sense to mark it `internal`?

src/core/Wyam.Configuration/ConfigScript/ScriptManager.cs
```diff
@@ -8,12 +8,26 @@
...
 using Microsoft.CodeAnalysis.Text;
+
+using Wyam.Common.Execution;
 using Wyam.Common.Tracing;
 using Wyam.Core.Execution;
 
 namespace Wyam.Configuration.ConfigScript
 {
-    internal class ScriptManager
+    internal interface IScriptManager
```
daveaglick: +1 for making an interface to make testing the `CacheManager` easier. Let's move it to a `IScriptManager.cs` file.

src/core/Wyam.Configuration/ConfigScript/CacheManager.cs
```diff
@@ -0,0 +1,127 @@
...
+            if (ConfigHashPath == null || ConfigDllPath == null)
+            {
+                return;
+            }
+
+            IFile configHashFile = _engine.FileSystem.GetRootFile(ConfigHashPath);
+            configHashFile.WriteAllText(scriptHash);
+
+            IFile configDllFile = _engine.FileSystem.GetRootFile(ConfigDllPath);
+            using (MemoryStream memory = new MemoryStream(_scriptManager.RawAssembly))
```
daveaglick, line 95: Can we add braces to this outer using - it's a nit, but I'm kind of particular about style :smile:

src/clients/Wyam/Commands/BuildCommand.cs
```diff
@@ -68,8 +68,9 @@ protected override void ParseOptions(ArgumentSyntax syntax)
...
             syntax.DefineOption("packages-path", ref _configOptions.PackagesPath, DirectoryPathFromArg, "The packages path to use (only if use-local is true).");
-            syntax.DefineOption("output-script", ref _configOptions.OutputScript, "Outputs the config script after it's been processed for further debugging.");
-            syntax.DefineOption("verify-config", ref _verifyConfig, false, "Compile the configuration but do not execute.");
+            syntax.DefineOption("output-script", ref _configOptions.OutputScript, "Outputs the config script after it's been processed for further debugging. Consider using the --ignore-config-hash directive when using this option.");
+            syntax.DefineOption("ignore-config-hash", ref _configOptions.IgnoreConfigHash, "Force evaluating the configuration file, even when no changes were detected.");
+            syntax.DefineOption("verify-config", ref _verifyConfig, false, "Compile the configuration but do not execute. Consider using the --ignore-config-hash directive when using this option.");
```
daveaglick: I wonder if either/both `output-script` and/or `verify-config` should implicitly turn on `ignore-config-hash` - not sure either of those options make sense if `ignore-config-hash` is off. In that case, we should note it in the help text and throw if `ignore-config-hash` is explicitly set to `false` when using one of the other two options. Thoughts?

src/clients/Wyam/Commands/ConfigOptions.cs
```diff
@@ -13,6 +13,7 @@ internal class ConfigOptions
         public bool UseGlobalSources = false;
         public DirectoryPath PackagesPath = null;
         public bool OutputScript = false;
```
daveaglick, line 15: We should probably add an extra option (with a corresponding CLI flag) like `OutputConfigAssembly` with a default value of `true` to allow opt-out of caching the assembly if the user wants to.

And if we're not writing the assembly, we probably shouldn't write the hash either (otherwise a previously saved assembly cache and the new hash may be inconsistent).

tests/core/Wyam.Configuration.Tests/ConfigScript/CacheManagerFixture.cs
```diff
@@ -0,0 +1,170 @@
...
+using Wyam.Common.IO;
+using Wyam.Configuration.ConfigScript;
+using Wyam.Testing;
+using Wyam.Testing.IO;
+
+namespace Wyam.Configuration.Tests.ConfigScript
+{
+    [TestFixture]
+    [Parallelizable(ParallelScope.Self | ParallelScope.Children)]
+    public class CacheManagerFixture : BaseFixture
```
daveaglick, line 22: Thanks for adding tests! How hard would it be to use test implementations instead of generated mocks? If the answer is "hard", then leave the tests as-is. Otherwise, I've been leaning towards full test object interface implementations over mocks recently.

daveaglick (changes requested):
Thanks a lot! Everything looks great, though I do have some comments/questions...

src/core/Wyam.Configuration/ConfigScript/CacheManager.cs
Silvenga: Sounds good.

src/core/Wyam.Configuration/ConfigScript/CacheManager.cs
Silvenga: Of course, I'll see if I can add the to R#'s settings.

src/core/Wyam.Configuration/ConfigScript/ScriptManager.cs
Silvenga: I was wondering about that, I'll move it out.

tests/core/Wyam.Configuration.Tests/ConfigScript/CacheManagerFixture.cs
Silvenga: Yeah, I was considering both, but I was having trouble figuring out how to perform cleanup/setup in isolation (file system setup/teardown) and I didn't want the tests to take 4 seconds to run.

NCrunch might be the other reason (it's not happy with me right now). 😹

src/clients/Wyam/Commands/BuildCommand.cs
Silvenga: I normally like explicit with cli interfaces, reduces the chance for unintended behavior - no one reads  cli docs.

Silvenga:
I think I got everything, rebased onto develop.

daveaglick:
WOOT! Thanks a ton for this one.

daveaglick (approved):

PR #559, opened 2017-08-27
Disable Data Protection Services

Silvenga:
I was trying to find a way to optimize razor compilation, it looked like there was a pause right before these messages are created, slowing down the compilation process during every step:

> Microsoft.Extensions.DependencyInjection.DataProtectionServices: User profile is available. Using 'C:\Users\User\AppData\Local\ASP.NET\DataProtection-Keys' as key repository and Windows DPAPI to encrypt keys at rest.

It turns out it probably doesn't, although removing it might slightly reduce compilation times (as the configuration heuristics run every time a module is declared). Wyam should hopefully never need to use DataProtection services and the messages are kind of useless to a user.

This pull disables the automatic DataProtection configuration heuristics from AspNetCore - which removes the above message.

### DIFF
--- a/src/extensions/Wyam.Razor/RazorCompiler.cs
+++ b/src/extensions/Wyam.Razor/RazorCompiler.cs
@@ -6,6 +6,11 @@ using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
 using System.Text.Encodings.Web;
+using System.Threading;
+using System.Xml.Linq;
+
+using Microsoft.AspNetCore.DataProtection.KeyManagement;
+using Microsoft.AspNetCore.DataProtection.Repositories;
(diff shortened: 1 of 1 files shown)


daveaglick:
Even if there's no performance improvement, I'm all for eliminating messages and features that don't make sense in a static generation context. Thanks.

I haven't taken too close a look yet, but the ASP.NET folks are rewriting parts of the Razor engine to presumably make it much more independent and easier to embed. It'll be interesting to see what the new stuff looks like compared to what we've got now. Planning to take a look at that after the .NET Core port. Hopefully you'll see some nice performance gains once it doesn't require basically all of ASP.NET just to compile and evaluate a Razor page.

PR #560, opened 2017-08-28
modify glob pattern to include subdirectories

archnaut:
Markdown files within sub-folders of the input blog folder did not have corresponding HTML files in output. 

Setting DocsKeys.BlogPath to glob pattern matching files within subfloders causes BlogIndex pipeline to fail on Razor module. It is because the _BlogIndex.cshtml is expecting a path to a folder not a glob. 

Modifying the glob pattern passed to the  ReadFiles module within BlogPosts.cs will allow subfolders to be searched for .md files without negatively impacting the BlogIndex pipeline.

### DIFF
--- a/src/recipes/Wyam.Web/Pipelines/BlogPosts.cs
+++ b/src/recipes/Wyam.Web/Pipelines/BlogPosts.cs
@@ -69,7 +69,7 @@ namespace Wyam.Web.Pipelines
                 MarkdownPosts,
                 new ModuleCollection
                 {
-                    new ReadFiles(ctx => $"{settings.PostsPath.Invoke<string>(ctx)}/{{!_,}}*.md"),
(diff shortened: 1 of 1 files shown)


daveaglick:
Thanks! This won't break anything so I've got no problems merging it, but I'm curious what the use case is. Given that the blog posts will still all be output directly under the `DocsKeys.BlogPath` folder (unless `DocsKeys.IncludeDateInPostPath` is specified in which case they'll be in date-based folders), why would blog posts be in child folders in the input? Is it to make organizing them easier?

archnaut:
Exactly, we have multiple contributors to blogs and we will have a lot of content on various topics so we need the ability to organize it. 

Thanks!

daveaglick:
That makes sense, happy to bring this in. Next release should be in a few days so it'll make it out then. Thanks for the contribution!

PR #577, opened 2017-09-19
AnalyzeCSharp Syntax: More than one C# attribute

jonasdoerr:
Fixed the behaviour that two attributes in the same line, e.g. `[Foo, Bar]`, result in duplicate syntax.


daveaglick:
Thanks!

Issue #395, opened 2017-01-19
Split out front matter delimiter logic into Wyam.Common as a public helper


daveaglick:
Will allow other tools to do similar front matter delimiting without pulling in the full `Wyam.Core` library. Needed initially for PoshWyam.

mbrameld:
I'd like to take this one on. Here's what I think should happen, am I on the right track?

Create a Split module in `Wyam.Common` with parameters for a start and end delimiter, at least one is required with start of file and end of file being defaults, respectively. 

Modify `Wyam.Blog` and `Wyam.Web` to use the new Split module.

Annotate the `FrontMatter` class as obsolete.

If that sounds reasonable I'll get started!

daveaglick:
@mbrameld Awesome, thanks! I think the idea here isn't to deprecate or replace the current `FrontMatter` module, but rather to move the bulk of its logic into general-purpose helper methods in Wyam.Common. That way any other third-party module or tool can do similar front matter processing without needing to reference the Wyam.Core library which has a number of extra dependencies.

Specifically, I see this as the following tasks:
* Taking the code at https://github.com/Wyamio/Wyam/blob/develop/src/core/Wyam.Core/Modules/Control/FrontMatter.cs#L112-L141 and extracting it
* Making the code not dependent on `IDocument` and use a `Stream` instead
* Putting the newly generic code into helper method(s) in Wyam.Common
* Changing `FrontMatter` to use the new general-purpose logic

mbrameld:
Got it, thanks!

Issue #34, opened 2015-06-19
Create a Split module


daveaglick:
Takes input document content and splits it into multiple documents based on a delimiter.

daveaglick:
Note that this could use the common delimiter splitting logic that's being moved to Wyam.Common in #395

daveaglick:
For example, the following input content:

```
A
---
B
---
C
```

Would result in three output documents with the content "A", "B", and "C" respectively.

PR #588, opened 2017-11-04
Extract splitter.

mbrameld:
Just want to make sure I'm on the right track with this for #395. It still needs to be abstracted further, documented, and tested, but is this along the lines of what you were thinking?

### DIFF
--- a/src/core/Wyam.Common/Util/Splitter.cs
+++ b/src/core/Wyam.Common/Util/Splitter.cs
@@ -1,0 +1,80 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Wyam.Common.Util
+{
+    /// <summary>
+    /// Splits a string based on a delimiter.
+    /// </summary>
+    /// <remarks>
+    /// This class is used to split some content on a deliimiter.
+    /// It is extracted from the FrontMatter module.
+    /// </remarks>
+    public class Splitter
+    {
+        private readonly string _delimiter;
+        private readonly bool _repeated;
+        private bool _ignoreDelimiterOnFirstLine = true;
+
+        /// <summary>
+        /// Constructs a Splitter instance with the specified delimiter.
+        /// </summary>
+        /// <param name="delimiter">The delimiter to use.</param>
+        /// <param name="repeated">True if the delimiter will be repeated. Defaults to false.</param>
+        public Splitter(string delimiter, bool repeated = false)
+        {
+            _delimiter = delimiter;
+            _repeated = repeated;
+        }
+
+        /// <summary>
+        /// Ignores the delimiter if it appears on the first line. This is useful when processing Jekyll style front matter that
+        /// has the delimiter both above and below the front matter content. The default behavior is <c>true</c>.
+        /// </summary>
+        /// <param name="ignore">If set to <c>true</c>, ignore the delimiter if it appears on the first line.</param>
+        /// <returns>The current Splitter instance.</returns>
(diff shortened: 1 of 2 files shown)


daveaglick:
Thanks - I can't believe it's been a month. Time flies. My apologies for the delay, hopefully you're still interested in finishing it out. I'm hoping to get some OSS time this weekend or early next week to review.

mbrameld:
Definitely, no sweat! I have a feeling it's not what you had in mind, but it's a starting point for discussion at least.

src/core/Wyam.Common/Util/Splitter.cs
```diff
@@ -0,0 +1,80 @@
...
+            _ignoreDelimiterOnFirstLine = ignore;
+            return this;
+        }
+
+        /// <summary>
+        /// Split the input.
+        /// </summary>
+        /// <param name="input">The input to split.</param>
+        /// <returns>A list of the split sections.</returns>
+        public List<string> Split(string input)
```
daveaglick, line 48: Wondering if we should return something other than a `List<string>` since it's always going to have either 1 or 2 items. 

One option would be to return the front matter content (or `null` if there isn't any) and then use a `ref` argument for the input and post-split content, that way if we didn't find the delimiter we could just leave the ref untouched: `string Split(ref string content) { ... }`

Another option would be to return a tuple: `(string frontMatter, string content) Split(string input) { ... }`

Thoughts?

src/core/Wyam.Core/Modules/Control/FrontMatter.cs
```diff
@@ -109,30 +107,17 @@ public override IEnumerable<IDocument> Execute(IReadOnlyList<IDocument> inputs,
...
-                    });
-                }
-                if (delimiterLine != -1)
+                Splitter splitter =
+                    new Splitter(_delimiter, _repeated)
+                        .IgnoreDelimiterOnFirstLine(_ignoreDelimiterOnFirstLine);
+
+                List<string> sections = splitter.Split(input.Content);
+
+                if (sections.Count == 2)
```
daveaglick, line 116: See the comment above about the `Split()` return value - testing for a specific number of items here doesn't quite feel right.

daveaglick (commented):
Finally had a chance to review. You've totally got the idea, only had one comment. As far as tests go, the FrontMatter module has quite a few - those can just be moved over and re-purposed for the new `Splitter` utility. Nice work!

daveaglick:
Actually, now that I think about it, I take back my comments in the code review. To make the new `Splitter` utility more general-purpose the `List<string>` return is probably the right way to go. For example, it could be used to drive a `Split` module (#34) that creates new documents from a delimiter regardless of how many times the delimiter appears. To support that scenario I suggest modifying the `Splitter` utility with a flag that indicates if the split should stop after it finds the first delimiter instance (as with the `FrontMatter` module) or continues splitting on all delimiter instances (as with the `Split` module).

Does that make sense?

daveaglick:
One other thought I just had is whether the `Split()` logic could be re-written to use a `Stream` (and then probably use a `StreamReader` internally) instead of an input string?

That would avoid us having to call `input.Content` on the input document. Since content for input documents is actually stored as a `Stream` internally, the `input.Content` call just ends up wasting resources by reading the whole document stream into a string. Since finding a delimiter(s) is a forward-only read operation, it should be possible to do it directly on a stream.

mbrameld:
I'll have a chance to knock this out on the first. Have a happy New Year!

PR #593, opened 2017-11-23
Fixes Output Corruption When Using No Clean

Silvenga:
Consider the following case:

- The `noclean` directive is enabled.
- File `a` is written with the content `12345`.
- Wyam re-executes the pipeline.
- File `a` is written with the content `abc`.

In the above case, file `a` should have the content `abc`, however, at current, the file will have the content `abc45`. This is not intuitive to the end user as it is assumed that when the `noclean` directive is set, that files should be completely overridden while the output directory is not removed.

This pull does the following:
- WriteFiles will now truncate the output file when the output content is less than the size of the file (when append is disabled).
- Fixes the testing fixtures to reuse the original file content under test (rather than starting with a empty file on rewrites).
- Configures the test runner NCrunch for solution support. 

The following are breaking changes:
- Some modules may assume that WriteFiles will not truncate.

### DIFF
--- a/Wyam.v3.ncrunchsolution
+++ b/Wyam.v3.ncrunchsolution
@@ -1,0 +1,6 @@
+﻿<SolutionConfiguration>
+  <Settings>
+    <AllowParallelTestExecution>True</AllowParallelTestExecution>
+    <SolutionConfigured>True</SolutionConfigured>
+  </Settings>
+</SolutionConfiguration>
--- a/src/core/Wyam.Configuration/Wyam.Configuration.v3.ncrunchproject
+++ b/src/core/Wyam.Configuration/Wyam.Configuration.v3.ncrunchproject
@@ -1,0 +1,7 @@
+﻿<ProjectConfiguration>
+  <Settings>
+    <AdditionalFilesToIncludeForProject>
+      <Value>..\..\..\themes\**.*</Value>
+    </AdditionalFilesToIncludeForProject>
+  </Settings>
+</ProjectConfiguration>
--- a/src/core/Wyam.Core/Modules/IO/WriteFiles.cs
+++ b/src/core/Wyam.Core/Modules/IO/WriteFiles.cs
@@ -323,6 +323,10 @@ namespace Wyam.Core.Modules.IO
                         using (Stream outputStream = _append ? output.OpenAppend() : output.OpenWrite())
                         {
                             inputStream.CopyTo(outputStream);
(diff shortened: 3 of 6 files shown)


daveaglick:
Thanks - I agree with you on expected behavior. I haven't had a chance to review yet, but I'll make sure we get this one in the next release (possibly early next week). Just a heads up that I haven't forgotten.

daveaglick:
> the next release (possibly early next week)

Lol. Best laid plans. What was I thinking with holidays coming up?

Everything looks good to so I'm going to merge. Thanks a lot!

FYI - I'm probably going to strip the NCrunch files back out and add NCrunch files to the `.gitignore`. They strike me as kind of like Resharper and other optional tooling configuration in that they probably shouldn't be in source control.

Silvenga:
> FYI - I'm probably going to strip the NCrunch files back out and add NCrunch files to the .gitignore. They strike me as kind of like Resharper and other optional tooling configuration in that they probably shouldn't be in source control.

@daveaglick Those files are designed to be in source control. It describes how to setup testing if the projects aren't isolated. I have to fix it everytime I open the project. 🐱

daveaglick:
> I have to fix it everytime I open the project.

Ugh. Do you have to fix it when you clone or every time you open the project from an existing local repo? If the latter, how does having them in source control help?

Regardless - having them shouldn't hurt anything for folks who don't have or use NCrunch so I don't feel strongly about excluding them. If it reduces friction for folks like yourself who use it, then I'm happy to add them back in.

Silvenga:
> Ugh. Do you have to fix it when you clone or every time you open the project from an existing local repo? If the latter, how does having them in source control help?

@daveaglick Na, just when I clone a new (sorry for my exaggerations). Although, it is a little annoying to have to stash those files when I move around.

> Regardless - having them shouldn't hurt anything for folks who don't have or use NCrunch so I don't feel strongly about excluding them.

Either way works, I've just spent some time figuring out why NCrunch is unhappy which might save other contributors some time. I've had to do it a couple times since I work on multiple devices in different stages of "care".

That said, my changes to NCrunch's configs only fix some of the problems (NCrunch isn't fully happy yet). Adding them to the gitignore should fix the issues with having to stash between checkouts and won't confuse people when they try using NCrunch (since it won't fully work). Now that I think about it, I think opting to exclude them would be better for everyone. NCrunch doesn't have nearly the community as R# yet.

PR #601, opened 2018-01-01
Markdown UseExtensions(types) Broken

Silvenga:
The `UseExtensions` call within the Markdown module does not support multiple extensions. This is caused by the `AddIfNotAlready` method checking for the generic type added. Since the generic type is always of type `IMarkdownExtension` the method would fail to add any additional extensions.


daveaglick:
Clever solution, I like it

PR #602, opened 2018-01-01
Ability to Give Markdown Engine Extension Directly

Silvenga:
On my Wyam powered project, I need to give the Markdown engine some context from the Wyam pipeline for custom Markdown extensions (i.e. image placeholder sizing). This would require a shared instance between Wyam and the Markdown engine, however, Wyam currently isolates extension materialization with itself. This pull request adds the ability to give the Markdown engine an instance of an extension.


daveaglick:
Thanks, another great addition

PR #604, opened 2018-01-04
Clarify wording of WithOuterHtml comment

olevett:
I was a little confused by the WithOuterHtml docs, as generated on wyam.io and ended up digging in code to check my assumptions. This wording is more explicit, which I hope helps.


daveaglick:
Thanks! Documentation corrections may often be small, but they can make or break how successful someone is at using the thing.

PR #647, opened 2018-03-05
Working on .NET Core

pauldotknopf:
This pull request isn't complete. It will be used to document my progress and callout things that are left to be done.


pauldotknopf:
Removed ```RecyclableMemoryContentStreamFactory```, because it didn't support netstandard.

pauldotknopf:
Temporarily removed ```SqlDataAdapter``` support. It exists in latest ```System.Data.SqlClient``` package (4.4), but were are on 4.3. I don't want to update the nuget packages just yet.

pauldotknopf:
I had to remove YamlDotNet.Dynamic, at least temporarily.

https://github.com/aaubry/YamlDotNet.Dynamic/issues/6

pauldotknopf:
I am leaving Wyam.CodeAnalysis as is (and Wyam.Docs recipe) because there is a dependencies on ManagedEsent. This would be a larger effort than I am willing to do.

pauldotknopf:
Keeping Wyam.Images as-is, until we can find a proper way to do ImageProcessor stuff, on .NET Core.

pauldotknopf:
Leaving Wyam.Tables as-is, until we can find .NET Core alternatives to it's dependencies.

pauldotknopf:
Leaving Wyam.TextGeneration as-is, until Rant can support .NET Core.

pauldotknopf:
Wyam.Hosting will be updated to use SignalR.

I also recommend using traditional middleware, instead of the Owin middleware.

pauldotknopf:
@daveaglick,

I left the testing and building stuff targeting full framework. What this does is essentially allow me to use the core of Wyam on .NET Core. I embed it an API level.

Things left:

**core**
- [ ] Wyam.Hosting - Convert to netstandard. Currently depends on Owin and Fleck. Fleck should be replaced with SignalR.

**extensions**
- [ ] Wyam.CodeAnalysis - The dependencies here are very Windows-centric. This might be better to just leave as is.
- [ ] Wyam.Images - There are .NET Core libraries for image manipulation available. It would take a good chunk of work to migrate from ImageProcessor though.
- [ ] Wyam.Tables - The dependencies here don't support .NET Core.
- [ ] Wyam.TextGeneration - Rant doesn't support .NET Core. It should be easy for them to add it though.

**general**

- [ ] Update all NuGet packages - Many libraries are referencing the 4.3.* family. 4.4 is out. I don't know a lot about the Wyam.Configuration and it's Cake-ish stuff, and that makes me uncomfortable to just "update everything".
- [ ] Bring back Sql for NET Core - I made Wyam.Core/Modules/Metadata/Sql.cs only available on full framework. Once nuget packages are updated, you can bring it back. The 4.4 family has SqlDataAdapter in it.
- [ ] Memory stream pool - I removed the memory stream pool because it didn't support netstandard. I can ifdef it back in for full framework only if needed.
- [ ] netstandard2.0 - I was heavy-handed with the 2.0. Some extensions/libs could target a lower version, but I doubt it is necessary.
I think the NuGet packages should be updated before this pull request is merged.

Is there anything else that you'd want me to look at?

Also, I can squash this pull request if you like.

daveaglick:
Awesome! It'll probably be another day or two until I have a chance to look closely at this, but I've got a hackathon coming up on Thursday so should have plenty of time to dig in then.

pauldotknopf:
Np, I know you MVP's are currently partyin...  :)   Carry on!

daveaglick:
This is crazy awesome.

Some comments/questions:
* What about `RecyclableMemoryContentStreamFactory` wasn't .NET Standard compatible? I see the "Microsoft.IO.RecyclableMemoryStream" package is `netstandard1.4` so was there something about the factory specifically that was a problem?
* I'm fine with temporarily removing/disabling non-primary modules or module features while we port. If that's a huge deal for anyone they can just hold off on updates. The move to .NET Core is the whole theme of the 1.x cycle so we can do whatever it takes to get there.
* I'll take care of Wyam.CodeAnalysis - that one may end up being pretty tricky. Same with Wyam.Configuration - that'll be a challenge too since it does a bunch with dynamic assembly loading.
* .NET Standard 2.0 is a good target - even if we can target lower versions for a few modules, I think it's better to standardize the version across all modules.

I'm going to just go ahead and merge this to reset our baseline so I can start hacking without creating conflicts in this PR. I'm also not going to squash - you deserve the magic GitHub commit points :smile:, and it might be helpful to have individual commits for each of the module ports.

For the remainder of the day, I'll probably try to fix a few outstanding bugs after merging this and then get a release out. If I have time after that, I'll start working on porting Wyam.Configuration.

pauldotknopf:
Awesome!

As for the memory pool: https://www.nuget.org/packages/Microsoft.IO.RecyclableMemoryStream/

I'm not seeing netstandard1.4.

pauldotknopf:
Hmm...  Looking at the source, I see netstanard: https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream/blob/master/src/Microsoft.IO.RecyclableMemoryStream.csproj

But, it isn't in the NuGet package. Maybe they need to push another update?

daveaglick:
Looks like it might just be an issue with how the gallery is displaying it - here's the actual package:
2018-03-08_16h17_06
I'll take a closer look and bring it back in if I can.

daveaglick:
Confirmed - brought it back, no problem.

Roblinde:
You guys rock! I'll make porting the Contentful module to netstandard my top prio once Wyam.Common is available.

PR #655, opened 2018-04-12
Fixed issue with month and day names in short/long date strings.

jonasdoerr:
There was a problem in `ToShortDateString()` and `ToLongDateString()`. The culture was picked correctly, but only `ShortDatePattern` and `LongDatePattern` were used. The culture itself was not passed. Therefore, the *current* culture was used which led to wrong day and month names.


daveaglick:
That’s a great catch, thanks for fixing it! And double thanks for adding tests.

jonasdoerr:
@daveaglick Do you mind if I add a `ToString(DateTime, IExecutionContext, string)` for date and time output?

daveaglick:
I don’t mind at all!

PR #658, opened 2018-04-29
OpenApi and Handlebars rendering - Not ready to merge at all

jrouaix:
Hello @daveaglick,
i'm on the way to add OpenApi reading & Handlebars rendering and have multiple question. So this PR is mostly about documenting questions.


jrouaix:
So 1st question : 
In the example I added I don't get the "myapi" metadata when I try to render handlebars files. I suspect it's because I didn't understand some pipeline works details.
What did I miss ?

[OK] I had to recreate the values sent to the handlebars template. If I just put the input.Metadata reference, it doesn't get all the values ... (don't really understand why yet)

jrouaix:
2nd one :
I added a "json" and "yaml" helpers in the handlebars renderer in order to debug my stuff. This could really be super helpfull but imo it really break the modularity of the framework because the handlebars part should not care about json & yaml serialisation, and other renderers, if they want to provide the same helpers, should rely on same code.
What is the good way to do that in your opinion ? 
- play with metadata through an other module, then out in the renderers,
- having some common methods to call in some Wyam.Helpers assembly, and all the renderers should expose helpers access to this methods,
- something else ?

daveaglick:
This is awesome! I'll get back to you with answers to your questions very soon (just don't have time today).

daveaglick:
Thanks for keeping this updated! It’s still on my radar - been stuck in MSBuild land lately, hopefully I’ll be emerging from that soon.

jrouaix:
Hi @daveaglick again, it's not ready to merge, but I'm still motivated to push this feature and use it. My dream is to replace https://swagger.io/tools/swagger-codegen/. Perhaps if you have time to answer my questions we'll be able to produce a "OpenApiCodeGen" recipe :-D

daveaglick:
Nice! Glad to see you haven’t lost interest in this even though I’ve been a terrible maintainer and haven’t given it any love. I’ll try really, really hard to get you some feedback soon :)

jrouaix:
Lol, you nasty maintainer. Do not worry, it's really ok.
I was waiting to be sure to be able to execute Wyam on linux ... it's ok now isn't it ?

We now are looking at GraphQL. The existing code gen is running on nodejs with handlebar templating.

So in a soon future, to generate Swagger & GrahQL apis, we'll use java/moustache & nodejs/handlebar.
And we really are a C# team ...

Issue #662, opened 2018-05-18
No Way to Override Hard-Coded Source in SourceRepositoryProvider.cs


rpriest1260:
Our build system blocks access to particular NuGet sources.  We have our own trusted server that we use.  Therefore, the hard-coded DefaultSources list in SourceRepositoryProvider.cs presents a problem when trying to generate documentation. There seems to be no way to opt out of using **"https://api.nuget.org/v3/index.json".**

Can this hard-coded value be removed, please?
Either that, or provide a way to disable it?

I would think that if a user specifies, in there WyamSettings:

       NuGetSources = new List<string>() {"https://mytrustednuget/"},
       UseGlobalSources = false,

that **only** that specified source should be used. If those particular settings are not specified, then it should use the sources configured in the user's\machine's NuGet Settings.

Could you please make this edit? Or, I could submit it if you wish. I intend to do nothing more than comment that line out.

Thanks.

daveaglick:
Sorry it took so long to get to this issue. Pretty sure it's the same as #673 (and even if it's not, the fix I added for that one should definitely resolve this one).

In the next version you'll be able to do the following, which will hopefully resolve your issues:

```c#
Wyam(new WyamSettings
{
    NuGetSources = new List<string> 
    {
        "<path-to-nuget-sources>",
    },
    IgnoreDefaultSources = true
});
```

If it doesn't please do circle back around and we'll reopen this issue.

FWIW, Wyam is dropping all direct NuGet support in the forthcoming 2.x release (see #668). That's not to say NuGet packages won't still be involved, just that Wyam will no longer be responsible for handling their installation - MSBuild will do that going forward.

rpriest1260:
Great! Thanks!  Standing by for the new version.

PR #663, opened 2018-05-19
Fix to address issue 662 - https://github.com/Wyamio/Wyam/issues/662

rpriest1260:
There are build systems that block access to "outside" NuGet sources.
Therefore, the hard-coded DefaultSources list in
SourceRepositoryProvider.cs presents a problem when trying to generate
documentation. There seems to be no way to opt out of using
"https://api.nuget.org/v3/index.json".
This fix removes this hard-coded value and leaves the decision to the
user via their NuGet.config setup or their WyamSettings specifications
(setting of NugetSources and UseGlobalSources).

### DIFF
--- a/src/core/Wyam.Configuration/NuGet/SourceRepositoryProvider.cs
+++ b/src/core/Wyam.Configuration/NuGet/SourceRepositoryProvider.cs
@@ -18,11 +18,6 @@ namespace Wyam.Configuration.NuGet
     /// </summary>
     internal class SourceRepositoryProvider : ISourceRepositoryProvider
     {
-        private static readonly string[] DefaultSources =
-        {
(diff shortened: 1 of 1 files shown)


daveaglick:
Sorry, totally forgot that you submitted this PR. I've addressed this issue with a new `--ignore-default-sources` flag and a corresponding Wyam.Cake setting. The default gallery source will still be added by default, but that can be ignored with the flag. That should cover this use case. I don't want to spend too much time tweaking NuGet handling since Wyam isn't going to manage NuGet packages starting with 2.0 (see #668).

rpriest1260:
Fair enough.  Having a way to ignore it in the WyamSettings should work.

PR #666, opened 2018-05-25
Fix blog recipe documentation.

tbolon:
Default value for ShowInNavbar seems to be "true" (the aboutme page is displayed in the navigation bar). You need to set this property to false to actually hide a page from the navigation bar.


daveaglick:
:+1: thanks - docs fixes like this are almost more important than code fixes for ensuring a good experience for users

PR #670, opened 2018-05-30
Fix hardcoded "blog" path in BlogArchives & BlogYearlyArchives pipelines

tbolon:


daveaglick:
Thanks, this was a great catch!

Issue #300, opened 2016-05-23
Add support for .NET Core


daveaglick:
Converted to netstandard2.0/netcoreapp2.0:

**Clients**
* [x] Cake.Wyam

2.x client strategy is discussed in #668

**Core**
* [x] Wyam.Common
* [x] Wyam.Configuration (need to look closely at custom assembly resolution)
* [x] Wyam.Core
* [x] Wyam.Hosting (need to investigate LiveReload support in .NET Standard)
* [x] Wyam.Testing
* [x] Wyam.Testing.JavaScript

**Extensions**
* [x] Wyam.AmazonWebServices
* [x] Wyam.CodeAnalysis
* [x] Wyam.Feeds
* [x] Wyam.GitHub
* [x] Wyam.Highlight
* [x] Wyam.Html
* [x] Wyam.Images
* [x] Wyam.Json
* [x] Wyam.Less
* [x] Wyam.Markdown
* [x] Wyam.Minification
* [x] Wyam.Razor
* [x] Wyam.Sass
* [x] Wyam.SearchIndex
* [x] Wyam.Tables
* [x] Wyam.Xmp
* [x] Wyam.Yaml
* [x] Wyam.YouTube

**Recipes**
* [x] Wyam.Blog
* [x] Wyam.BookSite
* [x] Wyam.Docs
* [x] Wyam.Web

**Client Tests**
* [x] Cake.Wyam.Tests

**Core Tests**
* [x] Wyam.Common.Tests
* [x] Wyam.Configuration.Tests
* [x] Wyam.Core.Tests
* [x] Wyam.Hosting.Tests

**Extension Tests**
* [x] Wyam.CodeAnalysis.Tests
* [x] Wyam.Highlight.Tests
* [x] Wyam.Html.Tests
* [x] Wyam.Json.Tests
* [x] Wyam.Markdown.Tests
* [x] Wyam.Minification.Tests
* [x] Wyam.Razor.Tests
* [x] Wyam.Sass.Tests
* [x] Wyam.Tables.Tests
* [x] Wyam.Xmp.Tests
* [x] Wyam.Yaml.Tests
* [x] Wyam.YouTube.Tests

**Integration Tests**
* [ ] Wyam.Examples.Tests

gijswijs:
Just to be helpful: OctopusDeploy release a tool to help with migrating to Core:

https://icanhasdot.net/result?github=Wyamio~2FWyam

daveaglick:
[...]

PR #671, opened 2018-05-30
Replace ImageProcessor with ImageSharp in Wyam.Images

MisinformedDNA:
Converts to netstandard2.0

Contributes to #300 

**Do not merge** 

- All code is not ported as I couldn't always find an equivalent. I'm blocked on various pieces of code.
- The code is not tested as there are no tests and, honestly, I'm new here and don't know how yet.

Just wanted to help move us to .NET Standard by moving to the new, specified image processor. This is the beginning, but more work needs to be done... and I need help.

I'm also not an "image guy", so please give as much feedback as you can!


daveaglick:
Thanks a bunch for starting on this one. I'm taking a look to see if I can unblock some of the differences.

daveaglick:
Got pretty far on this. I ended up basically totally refactoring the module (though the old code and your fixes we very helpful). Hopefully I'll have something to show for it tomorrow - still need to add some tests.

daveaglick:
All done. There's some breaking changes for anyone using the module, though basically all the old functionality is still available one way or another.

I added some simple tests to be sure ImageSharp isn't crashing and the output paths are as expected, but I have no idea how to actually tests the output so that'll have to do for now. I'll be on the lookout for bug reports when this goes out.

JimBobSquarePants:
>I have no idea how to actually tests the output so that'll have to do for now. 

I would simply guard yourself against regression and trust us to handle anything more complicated.

Generate a set of reference images using the library now and if they look correct, use them to compare against in the future.

daveaglick:
Thanks, great advice @JimBobSquarePants! I’m always happy to make things someone else’s problem :smile:

PR #672, opened 2018-05-31
Replace Tabular with EPPlus and CsvHelper

MisinformedDNA:
Contributes to #300


src/extensions/Wyam.Tables/ExcelFile.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Linq;
+using OfficeOpenXml;
+
+namespace Wyam.Tables
+{
+    internal static class ExcelFile
+    {
+        public static IEnumerable<IEnumerable<string>> GetAllRecords(Stream stream, int sheetNumber = 0)
+        {
+            using (var excel = new ExcelPackage(stream))
```
daveaglick, line 12: Looked through the EPPlus source and it doesn't look like `ExcelPackage` closes external streams. That means we need to make sure the input document stream gets disposed from the call site (`IDocument.GetStream()` requires the stream be closed as soon as possible since it locks some resources). I'll fix it up, just FYI.

daveaglick:

daveaglick:
Excellent work, thanks!

Issue #674, opened 2018-06-02
Type link with generic types in the docs recipe


k94ll13nn3:
I have a couple of remarks regarding the "autolink" feature in the docs recipe (the links in the documentation pages):
- When using a known type inside a generic type (like `IEnumerable<Foo>` when `Foo` is a type from the API), no link is created for the type (`Foo`).
- When trying to link a generic type from the API, it does not work (`Foo<T>` does not link to the `Foo<T>` API page). After a quick look at the code, the text is encoded before trying to find a match, and the `<` and `>` are not matched since they are replaced: https://github.com/Wyamio/Wyam/blob/9673b57967ebbbf1a7388105551e7d409be262d8/src/extensions/Wyam.Html/AutoLink.cs#L182
So I changed the `Api` pipeline locally, by changing `name` to `WebUtility.HtmlEncode(name)` when adding a new type link in this line: https://github.com/Wyamio/Wyam/blob/9673b57967ebbbf1a7388105551e7d409be262d8/src/recipes/Wyam.Docs/Pipelines/Api.cs#L67
I got it almost working, but there is some (expected I think) cases where it does not work:
  - If I use a generic parameter different from the one in the code, it does not match: `Foo<U>` does not links to the `Foo<T>` API page
  - `Foo<T,U>` won't link to `Foo<T, U>` (no space before `U` in the first one)

Is it possible to modify the `Api` pipeline in `config.wyam` in order to have this modification in my site or is a code modification necessary ?

daveaglick:
> Is it possible to modify the Api pipeline in config.wyam in order to have this modification in my site or is a code modification necessary

Yes, you can modify any pipelines you want. It introduces a little risk though because you have to basically replace the pipeline (you can actually replace individual modules, but that gets really complicated, really fast). If it changes in a later version, your customized version will be out of date.

```
int apiIndex = Pipelines.IndexOf(Docs.Api);
Pipelines.Remove(Docs.Api);
Pipelines.Insert(apiIndex, Docs.Api,
    new If(...)
);
```

(There's an open issue #591 to add a `.Replace()` method so you don't have to do the get index. remove, insert dance above)

That should work in the near-term for you. However, I'd like to get this change in directly. What to submit a PR? If not, no problem, I can just copy the code changes in.

k94ll13nn3:
Yes, I can do a PR.

daveaglick:
[...]

PR #679, opened 2018-06-07
 Add generation of autolinks for embedded generic types

k94ll13nn3:
As I said in https://github.com/Wyamio/Wyam/issues/674#issuecomment-395141531, I'm not 100% happy with the solution but it works and does not seem to break things. I considered adding a check for `_matchOnlyWholeWord` (equal to `true`) and/or `_querySelector` (equal to `code`) to restrict the case to the specific recipe usage but I wasn't sure that it was a good idea.

Preview:

### DIFF
--- a/src/extensions/Wyam.Html/AutoLink.cs
+++ b/src/extensions/Wyam.Html/AutoLink.cs
@@ -177,9 +177,11 @@ namespace Wyam.Html
             });
         }
 
-        private bool ReplaceStrings(IText text, IDictionary<string, string> map, out string newText)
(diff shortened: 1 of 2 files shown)


daveaglick:
I like where you're heading with this. I'm not totally happy with treating `<` and `>` as special cases right in the `AutoLink` module, but I think I've got an alternate approach that allows configuring additional arbitrary word boundary characters in the module. Then that can be added to the docs recipe specifically for angle brackets. I'll merge this in to capture the work you've done and then make some modifications provided I get it to pass your new test.

Thanks!

daveaglick:
Thanks! Seems to be working great and will go out with the next release.

Issue #651, opened 2018-04-04
Missing Sitemap in Blog recipe


elvogel:
The Blog recipe(s) do not include the Sitemap feature. not a high priority feature, but it sure would be nice if it were included! :)

PR #680, opened 2018-06-09
Add Sitemap Pipeline to Blog recipe

schulz3000:
#651 

Add default Sitemap Pipeline to Blog recipe


daveaglick:
Nice work, thanks! Especially like that it's in the common pipelines library so we can add it to docs later (though that would probably be a huge site map file).

PR #681, opened 2018-06-18
fix #590

### LINKED ISSUE #590: Tag link is wrong when using subdirectory
https://github.com/Wyamio/Wyam/blob/5a9f5d5e0bfbf48aa3984046864be11982e5eb92/themes/Blog/CleanBlog/_Index.cshtml#L70

I am using  `/subdirectory `as described [here](https://wyam.io/docs/deployment/serving-from-a-subdirectory). Tags link seems to work wrong (`/tags` instead of `/subdir/tags`). It works well when I replace referenced line with: 
                    `<a href="@Context.GetLink("tags/index.html")">View All Tags &rarr;</a>`

schulz3000:
fix #590 in all themes
Additional:
- similar problem with RSS Feed link in Trophy Blog theme
- similar problem with pagind in Samson Docs theme

### DIFF
--- a/themes/Blog/CleanBlog/_Index.cshtml
+++ b/themes/Blog/CleanBlog/_Index.cshtml
@@ -67,7 +67,7 @@
             </div>
             <ul class="pager">
                 <li class="next">
-                    <a href="/tags">View All Tags &rarr;</a>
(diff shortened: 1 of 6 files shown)


daveaglick:
Nice fix - thanks especially for catching the problem in all related themes. That'll also help ensure it doesn't creep back in when copying code from the themes.

PR #683, opened 2018-07-14
Updating BUILDING.md VS solution location

jdecarli:
I found `Wyam.sln` located in in the root folder, not in `src\`


daveaglick:
Thanks! Good catch - I’m out this week, but I’ll merge this in when I get back.

PR #684, opened 2018-07-19
Removing Google Analytics code from Stellar theme

jdecarli:
Customizing Stellar theme I found out this... I'm pretty positive that this doesn't belong here.

Please, feel free to reject the PR if that's not the case


daveaglick:
Really good catch - it absolutely doesn't belong there. Might have come over from HTML5 UP, not really sure, but regardless glad you found it it and took it out.

PR #685, opened 2018-07-24
fix for 'unrecognized expression' syntax error exception in the Stellar blog theme

jdecarli:
There's an extremely annoying error on the Stellar theme. When running it thru `wyam.exe -p` the console reads:
```
Uncaught Error: Syntax error, unrecognized expression: /
    at Function.ga.error (jquery.min.js:2)
    at ga.tokenize (jquery.min.js:2)
    at ga.select (jquery.min.js:2)
    at Function.ga [as find] (jquery.min.js:2)
    at m.fn.init.find (jquery.min.js:2)
    at new m.fn.init (jquery.min.js:2)
    at m (jquery.min.js:2)
    at HTMLAnchorElement.<anonymous> (main.js:93)
    at Function.each (jquery.min.js:2)
    at m.fn.init.each (jquery.min.js:2)
```
I thought it was the google analytics code, but no. It happens to be a piece of code in the `main.js` file. so I fixed it. I don't see any negative side-effect.

Let me know if it's better to create an issue before this PR. I thought It was too simple and small to create one.

### DIFF
--- a/themes/Blog/Stellar/assets/js/main.js
+++ b/themes/Blog/Stellar/assets/js/main.js
@@ -89,7 +89,7 @@
 
 							var	$this = $(this),
 								id = $this.attr('href'),
-								$section = $(id);
+								$section = $(id.replace('/',''));
 
 							// No section for this link? Bail.
 								if ($section.length < 1)


daveaglick:
No issue necessary, happy to take PRs however they come :smile:.

Most of the theme is from HTML5 UP, so it's entirely possible there are features we're not using or JS code that's broken. In this case it looks like this code had to do with highlighting sections, which we're not even doing, so IMO it's fine to modify (or even remove). Thanks!

jdecarli:
good point... I didn't double-check if that piece of code was used, I was under the assumption that it has to be for something. On the original theme works fine though (because it doesn't use any blackslash in the routes), but Wyam need those backslashes for correctly routing.

Anyways, thanks for the information!

PR #698, opened 2018-08-11
Add handling for missing leading forward slash

TheBrambleShark:
If `DocsKeys.BaseEditUrl` in `config.wyam` is missing a forward slash at the end, a malformed edit address is produced.

For example,

`Settings[DocsKeys.BaseEditUrl] = "https://code.luzfaltex.com"`

Will result in a url like:

`https://code.luzfaltex.comlegal/privacy/index.md`

This adds a quick check for the existence of a `/` at the end of the url and appends it if absent.


daveaglick:
Nice catch, thanks for the fix!

PR #748, opened 2018-11-20
Update markdown lib and use LinkRoot setting

tbolon:


daveaglick:
I loved everything about this PR, nice work! I'm going to add a `MarkdownPrependLinkRoot` setting to the recipes as well to trigger this behavior when using one of them.

PR #760, opened 2018-12-17
Fix BookSiteKeys.BookImage references in the theme so they don't assume the root directory.

damccull:
The default theme references BookKeys.Image as a basic string, which works fine until you try to use it in conjunction with Keys.LinkRoot. In that case the configuration key either needs to have the entire path (including the Keys.LinkRoot portion) or else the theme needs modified to use Context.GetLink() so it prepends the LinkRoot.


daveaglick:
:thumbsup: Thanks!

PR #761, opened 2018-12-23
Allow children of If modules to be manipulated like Pipelines

ociaw:
Currently, the If module doesn't provide any way to access its child modules. This makes modifying pipelines using If or ConditionalPipeline very difficult, since you essentially have to recreate the If module and its children.
For example, adding a Minify step to the Blog recipe's TagIndex pipeline currently requires this:
```
Pipelines["TagIndex"].Replace
(
    0,
    new If
    (
        ctx => ctx.Documents[Blog.Tags].Any(),
        new ReadFiles("_Tags.cshtml"),
        new FrontMatter(new Wyam.Yaml.Yaml()),
        new Wyam.Razor.Razor()
            .IgnorePrefix(null)
            .WithLayout("/_Layout.cshtml"),

        new MinifyHtml(),

        new WriteFiles((doc, ctx) => "tags/index.html")
    ).WithoutUnmatchedDocuments()
);
```
Which is brittle to recipe changes. With this, we can simply do
```
((IReadOnlyList<IModuleList>)Pipelines["TagIndex"][0])[0].InsertBeforeFirst<WriteFiles>
(
    new MinifyHtml()
);
```
Though I'm not sure if I've taken the best approach. Accessing the If module from the pipeline requires some ugly casting (though it might be unavoidable) and having Condition inherit from ModuleList feels dirty. Would it be better for Condition to inherit from ContainerModule and to instead treat Conditions as child modules of If?

### DIFF
--- a/src/core/Wyam.Core/Modules/Control/If.cs
+++ b/src/core/Wyam.Core/Modules/Control/If.cs
@@ -1,10 +1,9 @@
-﻿using System;
+﻿using System.Collections;
 using System.Collections.Generic;
 using Wyam.Common.Configuration;
 using Wyam.Common.Documents;
 using Wyam.Common.Modules;
 using Wyam.Common.Execution;
-using Wyam.Common.Util;
 
 namespace Wyam.Core.Modules.Control
 {
@@ -17,7 +16,7 @@ namespace Wyam.Core.Modules.Control
     /// outputs without modification.
     /// </remarks>
     /// <category>Control</category>
-    public class If : IModule
+    public class If : IModule, IReadOnlyList<IModuleList>
     {
         private readonly List<Condition> _conditions = new List<Condition>();
         private bool _withoutUnmatchedDocuments;
@@ -138,7 +137,7 @@ namespace Wyam.Core.Modules.Control
                 // Run the modules on the documents that satisfy the predicate
                 if (matched.Count > 0)
                 {
-                    results.AddRange(context.Execute(condition.Modules, matched));
+                    results.AddRange(context.Execute(condition, matched));
                 }
 
                 // Continue with the documents that don't satisfy the predicate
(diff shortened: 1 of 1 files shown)


daveaglick:
Thanks for this! It's an excellent proposal and you've clearly understood what's going on under the hood. I also think you're on the right track, implementation-wise.

> having Condition inherit from ModuleList feels dirty

I actually think that's the exact right approach. Each condition is after all just a list of modules to execute if it's true.

> Would it be better for Condition to inherit from ContainerModule and to instead treat Conditions as child modules of If?

That's probably overly broad - `If` shouldn't be able to contain anything but `Condition` and `ContainerModule` implies other arbitrary module children. But I think you're on the right track thinking about exposing `Condition` directly.

My thinking is to make `Condition` public and clean up the API a bit:
* Add setters for the `ContextConfig` and `DocumentConfig` that set the other to `null` when set - that way you can even adjust the conditions after the fact
* Instead of using a delegate like `ctx => true` for the final else clause, set both `ContextConfig` and `DocumentConfig` to `null` in that case and add a new property named something like `IsElse` to make it easy to determine if the condition is the final else clause

Then we can make `If` inherit from `IList<Condition>` and you can even adjust or add new conditions after the fact.

> Accessing the If module from the pipeline requires some ugly casting (though it might be unavoidable)

This is an excellent point, and while it's unrelated to this PR it got me wondering if there's an easier way to do this kind of thing. I was surprised there wasn't `.GetFirst<TModule>()` and/or `.Get<TModule>(int index)` extensions for `IModuleList`. If we add those then we can do something like:

```
Pipelines["TagIndex"].GetFirst<If>()[0].InsertBeforeFirst<WriteFiles>
(
    new MinifyHtml()
);
```

...which reads a _little_ better to me. I'll create a new issue for `.Get...()` extensions on `IModuleList` and get those in for the next release. I'm also going to merge this in as-is and make the changes mentioned above while I'm thinking about them. Thanks again!!

Issue #718, opened 2018-10-09
Wide Inheritance Diagrams are too small to read easily


SilentSin:
Inheritance diagrams for classes with lots of children or parents such as my [AnimancerState](https://animancer.github.io/api/Animancer/AnimancerState/) class get scaled down too small to read (especially on small screens).

I can see several possible solutions:

- Put the diagram below the list instead of beside.
- Put it below only if it passes a certain width.
- Is the list even necessary? Apart from the namespace, everything else is shown in the diagram.

daveaglick:
I've been playing around with the graph layout and placement to see if there's a better way of handling very wide graphs and I just can't come up with anything.

> Put the diagram below the list instead of beside.
> Put it below only if it passes a certain width.

For the graphs where they're too wide to read, this only buys a little extra width and most are still too wide to read after dropping them down.

> Is the list even necessary? Apart from the namespace, everything else is shown in the diagram

True, but since some graphs are going to be too wide to read well regardless, the list provides a cleaner way to see all the relationships and links. Even if all graphs could be readable, I like the list as an alternate presentation. I don't think I want to get rid of it.

I also tried making the graph vertically based instead of horizontal since our depth is likely to grow less than our width. That works to make everything readable, but it results in a very long graph for types with lots of derivatives/implementations. 

My final thought was to keep the diagrams as-is - like you said, most of the information is replicated in the list, so the large graphs can just serve as a quick visual reference and you don't need to really read them. Then I integrated https://github.com/ariutta/svg-pan-zoom to add pan/zoom support in case someone really wants to examine the larger graphs. This seems to strike a good middle ground to me.

PR #765, opened 2019-01-11
[WIP] Reverse direction of UML arrows.

elipriaulx:
I noticed the UML diagrams have backward arrows in the RxUI documentation ( https://reactiveui.net/api/reactiveui/appsupportjsonsuspensiondriver/ ), which I believe are being composed by Wyam. I have never used Wyam before and have no idea what I am doing, but I suspect these are backwards. 

Any further direction or feedback appreciated ^^.

### DIFF
--- a/themes/Docs/Samson/Shared/Kind/_NamedType.cshtml
+++ b/themes/Docs/Samson/Shared/Kind/_NamedType.cshtml
@@ -115,7 +115,7 @@
 							{
 								for(int c = 0 ; c < derivedTypes.Count ; c++)
 								{
-									hierarchy.AppendLine($"\tType-->Derived{c}[\"{System.Net.WebUtility.HtmlEncode(derivedTypes[c].String(CodeAnalysisKeys.DisplayName))}\"]");							
+									hierarchy.AppendLine($"\tDerived{c}[\"{System.Net.WebUtility.HtmlEncode(derivedTypes[c].String(CodeAnalysisKeys.DisplayName))}\"]-->Type");							
 									if(derivedTypes[c].ContainsKey(Keys.WritePath))
 									{
 										hierarchy.AppendLine($"\tclick Derived{c} \"{(Context.GetLink(derivedTypes[c].FilePath("WritePath")))}\"");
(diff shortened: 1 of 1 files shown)


daveaglick:
You are absolutely right, they're backwards! Can't believe I never noticed that, and I even spent many, many hours in courses back in the day that used or were about UML - guess we know how well that sunk in.

Thanks for the PR - I'll take a look and make sure it doesn't break anything, but otherwise looks good on first glance.

daveaglick:
The PR helps with the arrow direction, but it seems to reverse the layout...

Before:
2019-01-11_13h46_15

After:
image

I think the graph direction needs to be updated too. I'll merge this in and then play with it a little more. Thanks for the contribution!

glennawatson:
I think line 89 needs to change "graph td" to "graph bt" to make it look correct.

daveaglick:
Yep - that did the trick. I'm also looking at integrating https://github.com/ariutta/svg-pan-zoom for the super-wide graphs (see #718).

daveaglick:
2019-01-11_14h59_17

elipriaulx:
I'll get set up to build/debug wyam over the next week for a future changes. 
What's the go with the arrow heads? Would it be at all possible to have a hollow 'implements' arrow head, like Visual Paradigm and all the diagrams in Martin Fowler's books?

Eg:
odmpk

daveaglick:
It's looking very good now, thanks!

2019-01-11_16h22_05

elipriaulx:
Cool! That makes sense now :D

daveaglick:
> What's the go with the arrow heads? Would it be at all possible to have a hollow 'implements' arrow head

I agree that's more correct, but it doesn't look like mermaid can support or style arrow heads right now (would love to be proven wrong though)

elipriaulx:
Thanks for your interest and feedback on this @daveaglick , and @glennawatson for your input. I have had a browse through the mermaid documentation, and opened an issue for this here: https://github.com/knsv/mermaid/issues/780

ghuntley:
😂 thanks folks

PR #769, opened 2019-01-17
Allow the If indexer to be accessed without casting

ociaw:
`If` implements IList<IfCondition>, but the indexer is explicitly implemented. So instead of `Pipelines["TagIndex"].GetFirst<If>()[0]`, it needs to be cast first: `((IList<IfCondition>)Pipelines["TagIndex"].GetFirst<If>())[0]`.

Since the other members of IList aren't explicitly implemented and the release notes indicate the first way is expected to work, I've changed it to implement `this` implicitly instead.


daveaglick:
Good catch, thanks! Not sure why I implemented the interface explicitly but the intent was what you described.

PR #771, opened 2019-01-20
Make the globber case insensitive if the current file system is

glennawatson:
This fixes issues such as having globs like "../wyamio/wyam/**/*.dll" if on windows it's also WyamIO for the folder.

The approach generates a temporary file guid, does both lower case/upper case versions of the file, if they both exist then it considers the file system case insensitive.

Also updated the Microsoft.Extensions.FileSystemGlobbing package.

### DIFF
--- a/src/core/Wyam.Core/IO/Globbing/Globber.cs
+++ b/src/core/Wyam.Core/IO/Globbing/Globber.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


daveaglick:
> The approach generates a temporary file guid, does both lower case/upper case versions of the file, if they both exist then it considers the file system case insensitive.

That's very clever. In reading through the associated Stack Overflow question a few comments struck me as relevant:
* The system might have a mix of case-sensitive and case-insensitive file systems (especially on Linux where mounting other file systems is common).
* An approach that checks a known existing file (such as the current assembly) can avoid creating temporary files (which seem to cause a whole class of problems of their own - I'm particularly worried about file system race conditions where the temp file gets created but doesn't register as existing until after the `.Exists()` call, which could really screw up our globbing behavior).

I wonder if we can solve all these problems by checking the directory we feed the globber (and caching the result for each directory). That's more checks, but we generally only glob from the root or input directories so it's not significantly more work. Not to mention we're saving the disk I/O by checking existing directories. And if the root directory of the glob doesn't already exist, it's a moot point since this check is just for the globber and it wouldn't return results for a non-existing directory in the first place.

I.e., something like:

```
if(
    Directory.Exists(directory.Path.FullPath.ToUpper())
    && Directory.Exists(directory.Path.FullPath.ToLower()))
{
    // Case insensitive
}
```

Thoughts?

glennawatson:
I guess that could fail on Linux for all numeric file names and directories is the only comment that came to mind

glennawatson:
You could check for that case and fall back to the other way is a potential way around it.

daveaglick:
That’s what I’m thinking too - I’ll merge this as-is and then tweak it a bit.

daveaglick:
Thanks!

PR #772, opened 2019-01-20
Adding a git attributes to make line endings consistent on all platforms

glennawatson:
I was getting all sorts of line ending issues on windows. 

This will make it LF from now on for most file types.

If you get any git complaining about wanting to check out files all the file you can follow this:

https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,27 +2,26 @@
 * text=auto
 
 # Text files that should be normalized to LF in odb.
-*.cs     text eol=lf diff=csharp
-*.xaml   text eol=lf
-*.config text eol=lf
-*.c      text eol=lf
-*.h      text eol=lf
-*.cpp    text eol=lf
-*.hpp    text eol=lf
+*.cs     text diff=csharp
+*.xaml   text
+*.config text
+*.c      text
+*.h      text
+*.cpp    text
+*.hpp    text
 
 *.sln     text eol=crlf
-*.csproj  text eol=lf
-*.vcxproj text eol=lf
+*.csproj  text
+*.vcxproj text
 
-*.md       text eol=lf
-*.tt       text eol=lf
-*.sh       text eol=lf
-*.ps1      text eol=lf
-*.cmd      text eol=lf
+*.md       text
+*.tt       text
+*.sh       text
+*.ps1      text
+*.cmd      text
 *.bat      text eol=crlf
-*.markdown text eol=lf
-*.msbuild  text eol=lf
-
+*.markdown text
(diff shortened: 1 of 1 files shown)


daveaglick:
Educate me a little on git and line endings. I've never really paid that much attention (probably a luxury of working primarily on Windows). Why do we want to force LF for all files? If I look around at other `.gitattributes` files in .NET repositories, most seem to follow a pattern of leaving `text=auto` without overrides:

https://github.com/dotnet/corefx/blob/master/.gitattributes
https://github.com/cake-build/cake/blob/develop/.gitattributes
https://github.com/aspnet/AspNetCore/blob/master/.gitattributes
https://github.com/dotnet/roslyn/blob/master/.gitattributes

My (possibly incorrect) understanding is that as long as the local git `core.autocrlf` setting is `true` then it will check out with CRLF and convert back to LF on commit. Is the problem here that many files have already been committed with CRLF and so this change forces them back to LF-only regardless of the `core.autocrlf` setting?

I'm certainly not opposed to bringing this in, and I've already got a branch with the change ready to go that also normalizes all the files to the new setting - I just want to understand the implications a little better since this is a personal knowledge gap for me.

glennawatson:
I changed the files to mark the files as text only (removed the LF) except for the CRLF on the sln files.

The other .gitattributes rely on the first line (which you been missing) 

> text = auto

I had to turn off core.autocrlf = false to get files into your repository without them going crazy. All your files seemed to be stored as CRLF in git. Without turning it off I was getting files in my checkins as half CRLF (your parts) and LF (my parts) when I was checking in due to the core.autocrlf setting.

This line `text=auto` forces the equivalent of `core.autocrlf` on checkins. So CRLF locally (on windows) and LF on other platforms, then always LF on checkin. At the moment a lot of your files are CRLF on git.

Git's default way of determining if a file is text or binary is to check the first 8000 bytes, if it finds the NUL it marks it as text. We are just making it explicit for most of our common files.

In terms of the asplaintext this converts the file to text before doing a diff on the two revisions. So you can see the text contents of a doc/pdf file when doing a comparison between the two files when doing a 'git diff' on the msys platform.

daveaglick:
I get it - the problem is that both autocrlf isn't enforced _and_ that the files in the repo already have CRLF line endings, which screws up people that do have `autocrlf=auto`. I'll merge this without the enforced LF but then I'll submit a normalization commit to get all the line endings to LF-only so it works with auto from now on.

PR #773, opened 2019-01-20
Fix URL absolute link validation

glennawatson:
* Absolute validation would fail if it would fall through and do Get notification with a InvalidOperationException, now fixed to re-create the WebRequest. (not sure if Get checking is really needed in theory HEAD should always exist).
* Now using ValueTuple to avoid the heap allocation hit
* Was getting a bunch of mailto URL warnings. Now skip those since we don't want to do a webrequest against them.
* Was getting empty FilePath's, now using SourceString() to get guaranteed output
* The warnings inside the Validation methods (that we later add to the collection if there is failure) were spitting out Warning's rather than verbose, so we were getting a lot of duplication of warnings/error spitting out. Now using Verbose except for skipped unknown url types.

### DIFF
--- a/src/extensions/Wyam.Html/ValidateLinks.cs
+++ b/src/extensions/Wyam.Html/ValidateLinks.cs
@@ -1,9 +1,10 @@
-﻿using System;
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Net;
+using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
 using AngleSharp.Dom;
@@ -81,11 +82,11 @@ namespace Wyam.Html
 #pragma warning restore RCS1163 // Unused parameter.
 
             // Key = link, Value = source, tag HTML
-            ConcurrentDictionary<string, ConcurrentBag<Tuple<FilePath, string>>> links =
-                new ConcurrentDictionary<string, ConcurrentBag<Tuple<FilePath, string>>>();
+            ConcurrentDictionary<string, ConcurrentBag<(string documentSource, string outerHtml)>> links =
+                new ConcurrentDictionary<string, ConcurrentBag<(string documentSource, string outerHtml)>>();
 
             // Key = source, Value = tag HTML
-            ConcurrentDictionary<FilePath, ConcurrentBag<string>> failures = new ConcurrentDictionary<FilePath, ConcurrentBag<string>>();
+            ConcurrentDictionary<string, ConcurrentBag<string>> failures = new ConcurrentDictionary<string, ConcurrentBag<string>>();
 
             // Gather all links
             HtmlParser parser = new HtmlParser();
@@ -126,7 +127,7 @@ namespace Wyam.Html
                 int failureCount = failures.Sum(x => x.Value.Count);
                 string failureMessage = string.Join(
                     Environment.NewLine,
-                    failures.Select(x => $"{x.Key.FullPath}{Environment.NewLine} - {string.Join(Environment.NewLine + " - ", x.Value)}"));
+                    failures.Select(x => $"{x.Key}{Environment.NewLine} - {string.Join(Environment.NewLine + " - ", x.Value)}"));
                 Trace.TraceEvent(
                     _asError ? TraceEventType.Error : TraceEventType.Warning,
                     $"{failureCount} link validation failures:{Environment.NewLine}{failureMessage}");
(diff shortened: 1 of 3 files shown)


src/extensions/Wyam.Html/Wyam.Html.csproj
```diff
@@ -1,17 +1,18 @@
...
+    <RepositoryType>git</RepositoryType>
+  </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\core\Wyam.Common\Wyam.Common.csproj">
+      <PrivateAssets>all</PrivateAssets>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="AngleSharp" Version="0.9.10" />
+    <PackageReference Include="System.Net.Requests" Version="4.3.0" />
```
glennawatson: Added this so I can get some SourceLink magic when I was trying to diagnose the bug. With this package it now allows for debugging inside the requests.

glennawatson:
Also was finding a lot of false positives with CNAME redirections. Now do a absolute check if all the other relative links fail.

src/extensions/Wyam.Html/Wyam.Html.csproj
daveaglick: Great idea. The NuGet page for the `System.Net.Requests` package says
> Developers should prefer the classes in the System.Net.Http package

Would explicitly adding the `System.Net.Http` package instead accomplish the same SourceLink magic?

daveaglick:
I love everything about this PR, thanks for all the improvements to link validation :tada:. I'll merge as soon as we figure out `System.Net.Requests` vs. `System.Net.Http`.

glennawatson:
If you look at https://www.nuget.org/packages/System.Net.Requests/

System.Net.Requests includes 

```
.NETCore 5.0
System.Collections (>= 4.3.0)
System.Diagnostics.Debug (>= 4.3.0)
System.Diagnostics.Tracing (>= 4.3.0)
System.Globalization (>= 4.3.0)
System.IO (>= 4.3.0)
System.Net.Http (>= 4.3.0)
System.Net.Primitives (>= 4.3.0)
System.Net.WebHeaderCollection (>= 4.3.0)
System.Resources.ResourceManager (>= 4.3.0)
System.Runtime (>= 4.3.0)
System.Threading (>= 4.3.0)
System.Threading.Tasks (>= 4.3.0)
```

The inverse is not true for System.Net.Http

If you look at the top of the System.Net.Requests nuget description it mentions:
```
Commonly Used Types:
System.Net.HttpWebRequest
System.Net.HttpWebResponse
System.Net.HttpRequestHeader
System.Net.WebException
System.Net.WebExceptionStatus
System.Net.WebRequest
System.Net.WebResponse
```

While System.Net.Http is based around the newer HttpClient classes
```
Commonly Used Types:
System.Net.Http.HttpResponseMessage
System.Net.Http.DelegatingHandler
System.Net.Http.HttpRequestException
System.Net.Http.HttpClient
System.Net.Http.MultipartContent
System.Net.Http.Headers.HttpContentHeaders
System.Net.Http.HttpClientHandler
System.Net.Http.StreamContent
System.Net.Http.FormUrlEncodedContent
System.Net.Http.HttpMessageHandler
```

src/extensions/Wyam.Html/Wyam.Html.csproj
glennawatson: No, System.Net.Http does not contain the System.WebRequest's types, only HttpClient.

glennawatson: I was tempted to upgrade to HttpClient but decided against it.

glennawatson:
I can convert your validator pretty easily to use HttpClient, was considering going down that direction, but didn't want to make a change to working code.

daveaglick:
5db7cc7e-82d4-4dc2-b1d4-656efd1d4553

glennawatson:
Done @daveaglick

src/extensions/Wyam.Html/ValidateLinks.cs
```diff
@@ -29,6 +34,12 @@ namespace Wyam.Html
     /// <category>Input/Output</category>
     public class ValidateLinks : IModule
     {
+        private const int MaxAbsoluteLinkRetry = 5;
+        private const HttpStatusCode TooManyRequests = (HttpStatusCode)429;
```
glennawatson, line 38: HttpStatusCode doesn't have a member for 429 explicitly. Just a helper since I use it in a few spots.

daveaglick:
Looks fantastic, thanks a ton for all this! Should go out in a release very soon (hopefully tomorrow - don't quite have the bandwidth remaining today).

PR #816, opened 2019-04-21
Add documentation on how to use Include shortcode

johanvergeer:
As we discussed on [Stack Overlow](https://stackoverflow.com/questions/55548043/how-can-i-include-other-files-in-a-post-in-wyam/55592765?noredirect=1#comment97954862_55592765).

I have included some documentation on how to use the Include shortcodes to both the Include shortcode class and the Include module class. I have not been able to run a build yet, so I'm not sure whether the docs will turn out right.


daveaglick:
This is great, thanks! Merging it in (finally!) for a small catch-up release of Wyam and manually ported to Statiq.

PR #820, opened 2019-04-24
Add resolve NuGet using TargetFramework

devlead:
* This change will change package resolver to not use any moniker but instead target framework.

```diff
- Attempting to gather dependency information for package 'Wyam.Docs.2.2.4' with respect to project 'C:/projects/website-8e3xp/packages', targeting 'Any,Version=v0.0'
+ Attempting to gather dependency information for package 'Wyam.Docs.2.2.4' with respect to project 'C:/projects/website-8e3xp/packages', targeting '.NETCoreApp,Version=v2.1'
```

Also a bit out of scope, so could revert if you want

* Flattened exceptions when installing packages for error message without duplicates. ToString more complete error messages.
* Address potential async issue

### DIFF
--- a/src/core/Wyam.Configuration/NuGet/PackageInstaller.cs
+++ b/src/core/Wyam.Configuration/NuGet/PackageInstaller.cs
@@ -207,7 +207,7 @@ namespace Wyam.Configuration.NuGet
                     }
                     catch (Exception ex)
                     {
-                        Trace.Verbose($"Exception while installing packages: {(ex is AggregateException ? string.Join("; ", ((AggregateException)ex).InnerExceptions.Select(x => x.Message)) : ex.Message)}");
+                        Trace.Verbose($"Exception while installing packages: {(ex is AggregateException aex ? aex.Flatten() : ex)}");
                         Trace.Warning("Error while installing packages, attempting without remote repositories");
                         InstallPackages(packageManager, Array.Empty<SourceRepository>(), installedPackages);
                     }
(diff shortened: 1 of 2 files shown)


src/core/Wyam.Configuration/NuGet/PackageInstaller.cs
```diff
@@ -230,7 +230,7 @@ private void InstallPackages(NuGetPackageManager packageManager, IReadOnlyList<S
         {
             foreach (Package package in _packages.Values)
             {
-                package.Install(installationRepositories, installedPackages, packageManager).Wait();
+                package.Install(installationRepositories, installedPackages, packageManager).GetAwaiter().GetResult();
```
LokiMidgard, line 233: I know, unlike `Task.Result` this does not produce AggregateExceptions (as far as I know), so less nested exceptions. But the documentation states 

>[This API supports the product infrastructure and is not intended to be used directly from your code. [...] This type and its members are intended for use by the compiler.](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.taskawaiter.getresult?view=netframework-4.8#System_Runtime_CompilerServices_TaskAwaiter_GetResult)

Not that I against it. But it should be an informed decision.

daveaglick:
@devlead About to do a small release of Wyam to catch up on some of these pending PRs. Since Statiq no longer has direct NuGet support, this change would only affect this one release (which may end up being the final Wyam release). What benefit does this change provide? In other words, is it worth introducing changed behavior in the final release?

devlead:
It would use the right target framework when restoring.

daveaglick:
👍

src/core/Wyam.Configuration/NuGet/PackageInstaller.cs
daveaglick: I agree with the concern, thanks for calling it out. Normally I'd think carefully about using an internal API, but since the NuGet stuff is totally coming out in the next major version I'm just going to go ahead and YOLO merge this for now.

PR #855, opened 2019-06-17
Use latest version of Markdig

rchande:
We're exploring using Wyam in dotnet try, which also uses Markdig. Markdig made a breaking change by changing the signature of `MarkdownParser.Parse` to add an [optional parameter ](https://github.com/lunet-io/markdig/blob/ea6592b773def5e430b3af7c08ff774e9dc94c3e/src/Markdig/Parsers/MarkdownParser.cs#L74). dotnet try uses 0.17.0 so we get MissingMethodExceptions.

This change takes the latest release of markdig.

### DIFF
--- a/src/extensions/Wyam.Markdown/Wyam.Markdown.csproj
+++ b/src/extensions/Wyam.Markdown/Wyam.Markdown.csproj
@@ -12,6 +12,6 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Markdig" Version="0.15.4" />
+    <PackageReference Include="Markdig" Version="0.17.0" />
   </ItemGroup>
 </Project>


rchande:
Windows build is failing with the following; no clue what that means
```
EXEC : SignTool error : The specified PFX password is not correct. [D:\a\1\s\src\clients\Wyam\Wyam.csproj]
D:\a\1\s\src\clients\Wyam\Wyam.csproj(29,5): error MSB3073: The command ""C:/Program Files (x86)/Microsoft SDKs/Windows/v7.1A/Bin/SignTool.exe" sign /f "D:\a\1\s\src\clients\Wyam\..\..\..\digicert-davidglick.pfx" /p $(WYAM_CERTPASS) /fd SHA256 /td SHA256 /tr "http://timestamp.digicert.com" /d "Wyam" /du "https://wyam.io" "D:\a\1\s\src\clients\Wyam\bin\Release\netcoreapp2.1\Wyam.dll"" exited with code 1.
```

daveaglick:
Awesome!

You've managed to catch the project at an awkward time. I've been hard at work on v3 and it's different enough that I'm taking the opportunity to rebrand and relaunch everything as [Statiq](https://github.com/statiqdev). One of the fundamental changes, and one that I think will greatly benefit you, is a total separation of the _static site generator application_ (I.e., like Jekyll or Hugo) from the [underlying framework](https://github.com/statiqdev/Framework) (I.e., like ASPNET for static sites). The Roslyn-based scripting configuration is also gone and in it's place is much better support for using the framework from your own code. The result is something that is probably _much_ better suited to your use cases.

Which brings up a question: would it be better to use the existing Wyam release knowing that there are tons of breaking changes and a fundamental difference in usage on the horizon, or would it be better to start experimenting with the new stuff (which is still a little raw and will continue to break some, but getting closer to alphas)? I'm happy to jump on a chat or call and go into more details. If you want to keep playing with the current version, I'll get this merged and get a patch release out. If you'd rather start working with the new stuff, lets talk.

> Windows build is failing with the following; no clue what that means

Hmm - that's the code signing command. I recently rebuilt my Azure Pipelines to perform PR builds and I suspect the certificate password got lost along the way. I'll reset it, but otherwise the build looks good.

rchande:
@daveaglick Interesting. We're actually thinking about a couple of different use cases--might be good for us to do a quick call. How's your availability this week?

daveaglick:
@rchande Sounds good - sent you an email with my availability

daveaglick:
As discussed, going to get a Wyam update out the door today. Thanks!

PR #867, opened 2019-10-10
Moved scripts to fix render-blocking scripts

pug-pelle-p:
When running Light House on a new wyam project, it gives a warning that the js files blocks rendering. By moving the scripts files to the bottom of the page, we remove the render block.


daveaglick:
Cool, thanks! We can ignore the CI failure - looks like it's trying to sign the binary (which it shouldn't do in CI). Not going to spend a ton of time on that since I'm basically only doing small patches to Wyam at this point ([Statiq](https://github.com/statiqdev/Statiq.Framework) is where it's at).

Issue #878, opened 2019-12-13
Build: NuGet pack errors NU5125 (licenceUrl) and NU5048 (iconUrl)


tylerd:
URL locations for licence and icon are being deprecated in favor of embedding files inside the NuGet package.

[NuGet Warning NU5048](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5048)

[NuGet Warning NU5125](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5125)

This appears in 2 places during the build. 

**Create-Library-Packages** - Uses the `DotNetCorePack` cake function, which pulls the configuration from the `Directory.Build.props` file. This show up in the build as errors.

```xml
<PackageLicenseUrl>https://github.com/Wyamio/Wyam/blob/master/LICENSE</PackageLicenseUrl>
<PackageIconUrl>https://wyam.io/Content/images/logo-square-64.png</PackageIconUrl>
```

**Create-Theme-Packages** - Uses the `NuGetPack` cake function with a NuGetPackSettings input object. Only the NU5125 licenceUrl Warning, not the IconUrl.

```csharp
    IconUrl = new Uri("https://wyam.io/assets/img/logo-square-64.png"),
    LicenseUrl = new Uri("https://github.com/Wyamio/Wyam/blob/master/LICENSE"),
```

tylerd:
[Packing a license expression or a license file](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-a-license-expression-or-a-license-file)

[Packing an icon image file](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file)

daveaglick:
FYI - the original Wyam project is being moved to [statiqdev](https://github.com/statiqdev) to continue work there as _Statiq Web_ (the successor to Wyam and the Wyam blog recipe). Since this issue relates specifically to Wyam, I've transfered it to a new replacement project for legacy Wyam issues.

(This is an automated message)

PR #879, opened 2019-12-13
Updating Licence and Icon configuration for msbuild

tylerd:
Updating Licence and Icon configuration for msbuild in favor of embedding the files inside the NuGet package Issue: #878


tylerd:
@daveaglick it looks like the cert password configured in the pipeline is incorrect

PR #907, opened 2020-05-31
Sitemap inclusion metadata

drmathias:
Adds a setting to specify whether documents should be included in the sitemap. This enables certain documents to be excluded from the sitemap, such as data files or a 404 page.


daveaglick:
This is gold, you even added unit tests!

PR #909, opened 2020-06-06
Close <ItemGroup> tag in Statiq.Web.props

drmathias:
Seems to be causing the following error when building:

> error MSB4024: The imported project file "C:\Users\a\.nuget\packages\statiq.web\1.0.0-alpha.17\build\netcoreapp3.1\Statiq.Web.props" could not be loaded. The 'ItemGroup' start tag on line 2 position 4 does not match the end tag of 'Project'. Line 17, position 3.

### DIFF
--- a/src/Statiq.Web/Statiq.Web.props
+++ b/src/Statiq.Web/Statiq.Web.props
@@ -3,6 +3,7 @@
     <Compile Remove="theme\**" />
     <Compile Remove="extensions\**" />
     <Compile Remove="archetypes\**" />
+  </ItemGroup>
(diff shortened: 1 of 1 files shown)


daveaglick:
Oh crap! Don’t know how I missed that (and not exactly easy to unit test). Thanks for the catch! I’ll unlist the package and publish an update later tonight.

daveaglick:
Just published alpha.19 which resolves this problem and also makes sure the props flow transitively (there’s one in Statiq.App that should flow down). Thanks again for the catch!

PR #915, opened 2020-08-07
Add Statiq.Web .NET CLI template

devlead:
This adds a .NET CLI template, simplifies getting started with Statiq.Web

Currently just a project template, but nuget could contain more in the future.

Usage
```bash
dotnet new statiq-web -n MyWeb
```

Result
```
└───MyWeb
    │   MyWeb.csproj
    │   Program.cs
    │
    └───input
            index.md
```

When on nuget it would just be installed 
```
dotnet new -i Statiq.Web.Templates
```

To test pr
```bash
dotnet pack .\src\Statiq.Web.Templates\Statiq.Web.Templates.csproj
dotnet new -i .\src\Statiq.Web.Templates\bin\Debug\Statiq.Web.Templates.1.0.0-beta.3.nupkg
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,12 +9,23 @@ jobs:
   build:
     name: Build
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        working-directory: Statiq.Web
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
     steps:
       - name: Get Source
         uses: actions/checkout@v2
+        with:
+          path: Statiq.Web
+
+      - name: Checkout Statiq.Framework
+        uses: actions/checkout@v2
+        with:
+          repository: statiqdev/Statiq.Framework
+          path: Statiq.Framework
 
       - name: Install .NET Core SDK
         uses: actions/setup-dotnet@v1
--- a/src/Statiq.Web.Templates/Statiq.Web.Templates.csproj
+++ b/src/Statiq.Web.Templates/Statiq.Web.Templates.csproj
@@ -1,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <PackageType>Template</PackageType>
+    <PackageId>Statiq.Web.Templates</PackageId>
+    <Title>Statiq Web Templates</Title>
+    <Authors>Dave Glick</Authors>
+    <Description>Templates to use when creating a Statiq Web application, Statiq Web is a flexible static site generator.</Description>
+    <PackageTags>dotnet-new;templates;Statiq;Static;StaticContent;StaticSite;Blog;BlogEngine </PackageTags>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+
+    <IncludeContentInPack>true</IncludeContentInPack>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <IncludeSymbols>false</IncludeSymbols>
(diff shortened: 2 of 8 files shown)


devlead:
@daveaglick build error seems to be relative project references to Statiq. Framework which is in another repo.

daveaglick:
Thanks!!! This is great, anything to help onboarding is hugely appreciated.

Yeah, build errors are due to ongoing in-process increments between Framework and Web, need to figure out how to handle that at some point.

src/Statiq.Web.Templates/Statiq.Web.Templates.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <ContentTargetFolders>content</ContentTargetFolders>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="templates\**\*" Exclude="templates\**\bin\**;templates\**\obj\**;templates\**\.vs\**" />
+    <Compile Remove="**\*" />
+  </ItemGroup>
+
+  <Target Name="VersionBuild" BeforeTargets="PrepareForBuild">
+   <XmlPoke XmlInputPath="templates\statiq-web\StatiqWeb\StatiqWeb.csproj" Query="Project/ItemGroup/PackageReference/@Version" Value="$(Version)" />
```
devlead, line 25: This will ensure packed template references same package as built solution.

devlead:
> Yeah, build errors are due to ongoing in-process increments between Framework and Web, need to figure out how to handle that at some point.

Ok for now I've added so Statiq framework is cloned out too and now PR builds.

devlead:
@daveaglick rebased and conflicts sorted against latest main.

daveaglick:
Thanks! Hopefully I get to this one soon 🤞

src/Statiq.Web.Templates/Statiq.Web.Templates.csproj
daveaglick: 🎉 great! That way a new version of the template will just roll out with each release, right?

(or at least the build of the template will reference the currently releasing version, I'll probably need to add the last mile deployment of the template to the build script)

devlead: Correct it'll always be in sync with the version you're building.

.github/workflows/build.yml
```diff
@@ -9,12 +9,23 @@ jobs:
...
       - name: Get Source
         uses: actions/checkout@v2
+        with:
+          path: Statiq.Web
+
+      - name: Checkout Statiq.Framework
+        uses: actions/checkout@v2
+        with:
+          repository: statiqdev/Statiq.Framework
+          path: Statiq.Framework
```
daveaglick, line 28: Interesting, is the idea here that this'll make sure builds are successful even if Statiq Web relies on unreleased parts of Statiq Framework and has `<LocalReferences>true</LocalReferences>` in the root `Directory.Build.props` (as is often the case between releases). If so, this is awesome.

daveaglick: Saw the comment above which confirms, this is a great way to handle it for the Actions verification build since it's often in a state where it's using pre-release functionality.

daveaglick:
Thanks a bunch for this, it'll make onboarding a lot easier I think!

I can see expanding the set of templates as we go to include themes or other styles of sites. I think Statiq.Web is the right place for this too. I was considering a separate repo for templates earlier, but it's probably better to make sure they cycle versions and release with the specific project. We can create a separate Statiq.Docs.Templates later once that's mature enough.

PR #918, opened 2020-09-19
Fixed bug that prevented processing of non-ascii symbols

Vladekk:
context.Request.Path.ToString() was leaving url encoded symbols as is. It means when further code tried to check if file exists on disc, result was always false in case of non-latin filenames.

I am not sure if any additional code is neccessary here. For some reason, I can't open and run Statiq solution on my PC properly.


daveaglick:
Thanks a ton for running this down! What's a good filename I can use to test locally just to verify the fix for you?

Vladekk:
> Thanks a ton for running this down! What's a good filename I can use to test locally just to verify the fix for you?

Just use any cyrillic word, for example Тест, which means Test.

daveaglick:
Thanks again!

PR #970, opened 2021-11-04
Support modifying Feed paths

### LINKED ISSUE #960: Modifying RSS Feed Path
Currently I'm using `Statiq.Web` as-is with no crazy configurations. I've got my RSS feed generating (`feed.rss`) however I want to change the name to `feed.xml` but I'm struggling.

Looking at the source code, I see the `GenerateFeeds` module has the exact method I want but I'm having trouble getting access to the module. Following the code that calls it, I see it is in a nested `ExcuteConfig` inside a `ForEachDocument` module...

https://github.com/statiqdev/Statiq.Web/blob/d9da990fd75b14575d7679273373d7ab5da54b78/src/Statiq.Web/Pipelines/Feeds.cs#L24-L130

I see that the generate feeds module is returned in `ExecuteConfig` but while debugging my application, I can't find where the `GenerateFeeds` module ends up.

My thought was to add a post-process module (like the following) to do what I'm wanting:

```csharp
ModifyPipeline(
	nameof(Statiq.Web.Pipelines.Feeds),
	x.PostProcessModules.Add(new ForAllDocuments
	{
		new ExecuteConfig(Config.FromDocument(feedDoc =>
		{
			// somehow get the `GenerateFeeds` instance
			generateFeeds.WithRssPath(feedDoc.Destination.ChangeExtension("xml"));
		}))
	})
);
```

I'm open to completely different suggestions for doing the same type of thing though! Really I just want the RSS file named the same so I don't need to update it in a bunch of places, not because I have a preference over the file extension itself.

Turnerj:
Fixes #960 

Using `FeedRss`, `FeedAtom` and `FeedRdf` keys, you can specify `true`, `false` or a string path. Specifying true/false will behave as before. When you specify a path, this will be the destination path used for the feed itself.

### DIFF
--- a/src/Statiq.Web/Pipelines/Feeds.cs
+++ b/src/Statiq.Web/Pipelines/Feeds.cs
@@ -61,9 +61,9 @@ namespace Statiq.Web.Pipelines
                         // Generate the feed(s)
                         GenerateFeeds generateFeeds = new GenerateFeeds()
                             .MaximumItems(feedDoc.GetInt(WebKeys.FeedSize))
-                            .WithRssPath(feedDoc.GetBool(WebKeys.FeedRss, false) ? feedDoc.Destination.ChangeExtension("rss") : null)
-                            .WithAtomPath(feedDoc.GetBool(WebKeys.FeedAtom, false) ? feedDoc.Destination.ChangeExtension("atom") : null)
-                            .WithRdfPath(feedDoc.GetBool(WebKeys.FeedRdf, false) ? feedDoc.Destination.ChangeExtension("rdf") : null)
+                            .WithRssPath(TryGetFeedPath(feedDoc, WebKeys.FeedRss, "rss", out NormalizedPath feedPath) ? feedPath : null)
+                            .WithAtomPath(TryGetFeedPath(feedDoc, WebKeys.FeedAtom, "atom", out feedPath) ? feedPath : null)
+                            .WithRdfPath(TryGetFeedPath(feedDoc, WebKeys.FeedRdf, "rdf", out feedPath) ? feedPath : null)
                             .WithFeedId(feedDoc.GetString(WebKeys.FeedId))
                             .WithFeedTitle(feedDoc.GetString(WebKeys.FeedTitle))
                             .WithFeedDescription(feedDoc.GetString(WebKeys.FeedDescription))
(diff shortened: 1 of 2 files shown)


Turnerj:
FYI, the build errors seem to relate to your recent commit adding `MediaTypes.Statiq` but that isn't available in the NuGet version of Statiq.App. For myself, I saw the `Statiq.Web.csproj` file does support using a local `Statiq.Framework` but wasn't sure the way I was meant to enable it to do that locally (I just commented out those erroring lines and ran the new test and it passed).

daveaglick:
Yeah, I still haven't figured out a good way to run PR build checks while Web has dependencies on in-flight changes in Framework. I should probably figure out how to download the Framework source and use it for testing, but that's a problem for another day :)

src/Statiq.Web/Pipelines/Feeds.cs
```diff
@@ -139,5 +139,19 @@ public Feeds()
 
         public static bool IsFeed(IDocument document) =>
             document.ContainsKey(WebKeys.FeedRss) || document.ContainsKey(WebKeys.FeedAtom) || document.ContainsKey(WebKeys.FeedRdf);
+
+        private static bool TryGetFeedPath(IDocument feedDoc, string key, string defaultExtension, out NormalizedPath path)
```
daveaglick, line 143: Nice!

daveaglick:
Looks great, thanks!

src/Statiq.Web/Pipelines/Feeds.cs
Turnerj: Thanks! I was pretty happy with how that turned out!

Turnerj:
> 
> 
> Yeah, I still haven't figured out a good way to run PR build checks while Web has dependencies on in-flight changes in Framework. I should probably figure out how to download the Framework source and use it for testing, but that's a problem for another day :)

I know that the `checkout` action in GitHub Actions can clone other repos but depending whether you build your release packages via it or not, you'll likely run into issues about whether it references your `Statiq.Framework` NuGet package correctly or not.

(Also, just in case you didn't see it with all the notifications you've gotten from me in the last few days, I've got 2 small PRs on Statiq.Framework you might want to look at when you have time 😊)

PR #979, opened 2022-01-19
Also generates an excerpt for cshtml sources

rwwilden:
Because `cshtml` is also translated to html, makes sense to be able to generate excerpts from `cshtml` files as well. Added some tests to confirm that it works.

### DIFF
--- a/src/Statiq.Web/Pipelines/Content.cs
+++ b/src/Statiq.Web/Pipelines/Content.cs
@@ -27,7 +27,8 @@ namespace Statiq.Web.Pipelines
                         new OptimizeFileName()
                     },
                     new RenderContentProcessTemplates(templates),
-                    new ExecuteIf(Config.FromDocument(doc => doc.MediaTypeEquals(MediaTypes.Html)))
+                    new ExecuteIf(Config.FromDocument(
+                        doc => doc.MediaTypeEquals(MediaTypes.Html) || doc.MediaTypeEquals(MediaTypes.Razor)))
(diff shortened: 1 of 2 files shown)


rwwilden-rtl:
Now that I'm learning more about Statiq, it seems that excerpt generation takes place before the actual generation of the HTML. So for example, `@using` statements will show up in the excerpt, which wasn't entirely my intention.

So maybe this isn't such a good idea after all.

daveaglick:
> Now that I'm learning more about Statiq, it seems that excerpt generation takes place before the actual generation of the HTML.

Yeah, unfortunately I've been down this road before and the behavior when generating excerpts from anything from HTML content can be very unexpected. It's a classic chicken-and-egg problem: we want to use the excerpt inside a Razor page for something like a list of articles, but those articles need to be rendered before we can. If the ordering here was well-defined we _might_ be able to render certain files before other files, but there's no good way to determine that and in many cases it's even circular (think about a tag cloud on a blog post that itself has a tag). This is the entire reason we have a [process phase followed by a post-process phase](https://www.statiq.dev/framework/pipelines/#phases) - so that we can at least get HTML from _some_ of the files (I.e. Markdown) and do other processing before running through Razor.

So the short story is you're right, at the moment we can only reliably generate excerpts for documents that produce HTML content coming out of the process phase - basically Markdown and raw HTML. For others, it's perfectly suitable to manually define and excerpt using an `Excerpt` property in front matter.

PR #982, opened 2022-02-13
Moving lookup of document base href outside of loop

phil-scott-78:
We can grab the base URL once outside of the LINQ chain which starts to add up especially with some link heavy documents in the API.  This dropped the number of calls total on my run for `GetElementsByTagName` from 190,000 down to 9,000 and shaved a few seconds off the end of the runs.


daveaglick:
Love it! Big performance gains often come from an accumulation of little performance gains like this (especially given that this analyzer is particularly slow right now). Not to mention the reduction in memory pressure reducing the calls by more than an order of magnitude. Nice work.

PR #989, opened 2022-04-23
Fix validation if link contains non-ASCII characters

theranger:
Link validation fails if the link was generated from a file that contains non-ASCII characters. Although links themselves are created properly and are urlencoded, the validator misses decoding step.


daveaglick:
Sorry for the extremely late response here - I got swamped over the summer and am just now starting to dig out. It does look like a very similar change to this PR was already made along the way - there were a number of bug reports related to this behavior so I must have cleaned it up at some point. Thanks a ton for taking the time to run this down though - if nothing else your having solved it in the same way is a great validation of the fix!

PR #1000, opened 2022-10-06
Fix Lunr Search Resource Paths

girlpunk:
At present, setting the `SearchScriptPath` key does not change the script path as expected, but changes the index path. This patch fixes the `SearchScriptPath` to work as expected, as well as adding keys to set the path for the index and results file.

Note that as this changes the behaviour of the existing `SearchScriptPath` key, this should be considered a breaking change. However, dealing with this is simply a matter of changing `SearchScriptPath` to `SearchIndexPath` in any configuration.

This also addresses a "feature" in Azure Static Apps, where files ending in `.gz` are served with compression headers, causing browsers to automatically decompress the file before passing the result to javascript. This causes the search script to fail, as it is expecting compressed data.

### DIFF
--- a/src/Statiq.Web/Pipelines/SearchIndex.cs
+++ b/src/Statiq.Web/Pipelines/SearchIndex.cs
@@ -46,7 +46,9 @@ namespace Statiq.Web.Pipelines
 
                         // Create the module
                         GenerateLunrIndex generateLunrIndex = new GenerateLunrIndex()
-                            .WithIndexPath(ctx.GetPath(WebKeys.SearchScriptPath))
+                            .WithScriptPath(ctx.GetPath(WebKeys.SearchScriptPath, NormalizedPath.Null))
+                            .WithIndexPath(ctx.GetPath(WebKeys.SearchIndexPath, NormalizedPath.Null))
+                            .WithResultsPath(ctx.GetPath(WebKeys.SearchResultsPath, NormalizedPath.Null))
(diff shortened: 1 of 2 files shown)


girlpunk:
(Would it be possible to have the Hacktober tags added to this and my other PR https://github.com/statiqdev/Docable/pull/17? https://hacktoberfest.com/participation/#maintainers Thanks!)

daveaglick:
> Would it be possible to have the Hacktober tags added to this and my other PR

Yeah - sure thing, and thanks! I've been swamped with personal stuff lately (which seems to be the norm these days), but hopefully I'll be able to circle back here in a day or two. Just wanted to let you know I haven't forgotten or disappeared.

daveaglick:
> I'll be able to circle back here in a day or two

Best laid plans, am I right?

Taking a look at this today though. The repo does have the `hacktoberfest` topic, so you should be good to go there (assuming you haven't already met the requirements through other PRs).

Great catch on the use of search in Statiq Web too - thanks! I spent a lot of time on the underlying module, but apparently not so much integrating it here 😬.

If you don't mind signing the CLA (there was a bug in the CLA process, which is why it didn't show up right away), I'll get this merged.

girlpunk:
> Best laid plans, am I right?

For an open source project, you're doing better than most

> assuming you haven't already met the requirements through other PRs

Best laid plans strikes again

> If you don't mind signing the CLA

I think that should be done now

daveaglick:
Thanks again!

PR #1015, opened 2023-08-17
Use new Azure Search API

### LINKED ISSUE #1014: Microsoft.Azure.Search is Deprecated, and has known vulnerabilities
`Statiq.Web.Azure` depends on `Microsoft.Azure.Search`, which is deprecated and should no longer be used.

Additionally, `Microsoft.Azure.Search` has a dependency on `Microsoft.Rest.ClientRuntime`, which has [known vulnerabilities](https://github.com/advisories/GHSA-whph-446h-6m9v).

Microsoft's official recommendation is to use `Azure.Search.Documents` instead of `Microsoft.Azure.Search`.

girlpunk:
Fixes #1014 

Note that I haven't been able to fully test this, particularly the retry portions of Polly.

### DIFF
--- a/src/Statiq.Web.Azure/DeploySearchIndex.cs
+++ b/src/Statiq.Web.Azure/DeploySearchIndex.cs
@@ -1,16 +1,15 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
-using System.Net.Http;
-using System.Net.Http.Headers;
-using System.Text;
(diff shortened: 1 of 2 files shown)


daveaglick:
I know it's super old, but thanks for putting this together! It's been integrated into #1024 and merged there. Not sure if you're still using Statiq or not, but the hope is to consolidate all the projects and release a 1.0 version to NuGet this week.

PR #1022, opened 2024-09-08
Updated to .Net 8

NikoMix:
This Pull Request is updating all projects to target .net 8

However, it's dependant on https://github.com/statiqdev/Statiq.Framework/pull/278 and https://github.com/phmonte/Buildalyzer to release a new version (due to issues in resolving legacy dependencies).

Buildalyzer has already changes merged, but not yet released a version with a fix. For performing the update, I used a local reference - once the version is released I'll reverse that change, until then it's just a Draft Pull Request.

In addition Microsoft.Azure.Search had been deprecated (used bei Statiq.Web.Azure) and should be replaced by Azure.Search.Documents. Will migrate this, but did not yet have the chance to setup a test instance.

PR #1024, opened 2025-12-29
Update TargetFramework to net10.0 and bump all dependencies

jasonsummers:
## Credits
- @girlpunk for #1015 - included in this PR
- @neromix for #1022

## Summary
- Replaces `netcoreapp3.1` with `net10.0` as the target framework for the solution.
- Bumps all dependencies in the solution
- Removes Nuget package vulnerability suppressions.
- Adds suppressions for new warnings which would be time consuming to resolve or grow the size of the PR.
- Pins some transitive dependencies to remediate security issues.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,7 +22,7 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TargetFramework>net10.0</TargetFramework>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <WarningsNotAsErrors>NU1901;NU1902;NU1903;NU1904;CA1021;CA1062;CA1724;SA1414;</WarningsNotAsErrors>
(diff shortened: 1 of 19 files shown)


daveaglick:
Great timing! I've actually got some (rare) downtime this week and was considering doing some cleanup in here. What I'll probably do is YOLO merge this and the related PR in Statiq.Framework just to get the ball rolling. Then I'll *finally* collapse the repos, bringing Statiq.Framework into this one, archiving that one, and renaming this to just "Statiq". The idea is that I want to get everything down to a single monorepo, relicense it all to MIT, and prepare for a 1.0 release.

daveaglick:
Here goes nothing!

Thanks a ton for bringing this all together and wrapping it up in a single PR. I'll do plenty of testing locally since I also need to update the GitHub Actions configuration, but hopefully I'll have this out to NuGet by the end of the week.

